PR #18, opened 2016-02-13
Fix plurals (typos?).

mkcor:
Right?


tfmorris:
Thanks for your contribution!

PR #20, opened 2017-03-27
Fix markdown H1 in index.md

jezcope:
Since Jekyll (and hence GH pages) switched to using kramdown instead of redcarpet for markdown processing, headers *must* have a space between the last `#` and the header text.


magdmartin:
Thank you :)

PR #46, opened 2019-09-26
Add a simple note about data locality and privacy.

thadguidry:


wetneb:
Thanks, I think we need something like that since the question is quite frequently asked. I might rephrase and expand that paragraph into its own page, explaining which OR features do communicate with the internet (reconciliation, fetch urls, Google Sheets, database…).

thadguidry:
Yeah, the idea was to also eventually add a "Click to read more" which then points to a Wiki page dedicated to that information (Gustavo mentioned a few things on our mailing list)

thadguidry:
Ideally, on those features that expose data, we could add a confirmation popup with "This feature of OpenRefine will share share parts of your data which may be private for you...Are you sure you want to continue? (yes/no/always)"

PR #49, opened 2019-12-04
Removing download links for OpenRefine 3.1

thadguidry:
Also marked OpenRefine 3.1 as a pre-release in Github since we know now that it has some serious bugs.

### DIFF
--- a/download.md
+++ b/download.md
@@ -39,20 +39,6 @@ Download, open, drag icon into the Applications folder and double click on it.
 + **[Linux kit](https://github.com/OpenRefine/OpenRefine/releases/download/3.2/openrefine-linux-3.2.tar.gz)**, 
 Download, extract, then type _./refine_ to start.
 
-
-### OpenRefine 3.1
-The final release of OpenRefine 3.1. Please BACKUP your workspace directory before installing and report any problems that you encounter.
-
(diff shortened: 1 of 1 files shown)


wetneb:
Removing it from the downloads seems sensible to me but I would rather not retrospectively change the status of a release, that is likely to break some workflows I would say!

thadguidry:
I'd want those workflows against 3.1 to break...and someone ask us about it....so we can tell them why.  Given that, this is fine I think for 3.1.

But agree that for the future, this should not be necessary if we do due diligence towards testing in beta prior to releasing.  Having smaller incremental changes would be best to help test, but when its major sweeping changes (like what you did for Jackson migration), let's keep those major changes in beta for a much longer period top help all around.

wetneb:
I really think there is no benefit in retrospectively marking 3.1 as a pre-release: who would download this version specifically now? Instead of going for 3.2 directly? This is likely to break integrations, for instance with package managers which rely on GitHub releases.

thadguidry:
I think both of us are worrying about this too much, in hindsight.  Reverted "pre-release" tag on 3.1 version.

The lesson has already been learned for us and we'll commit to testing more thoroughly in the future.

wetneb:
Thanks!

PR #58, opened 2020-01-09
Add CS&S announcement

wetneb:
We still need to add URLs on the steering committee members.


_layouts/index.html
```diff
@@ -18,8 +18,13 @@ <h1 id="welcome">Welcome!</h1>
 (It works by running a small server on your computer and you use your web browser to interact with it)</p>
     
 <p>OpenRefine is available in English, Chinese, Spanish, French, Russian, Portuguese (Brazil), German, Japanese, Italian, Hungarian, Hebrew, Filipino, Cebuano, Tagalog </p>
-    
-OpenRefine was supported in 2018 by:
+
+OpenRefine is sponsored by the Code for Science & Society:
```
wetneb: I would not but CS&S and GNI on the same footing: the first is a fiscal sponsor, the second is a funder.

PR #65, opened 2020-03-07
add online course link

n1n9-jp:


trnstlntk (commented):
This addition was suggested in 2020, so it's a bit stale I'd say.

It links to a Japanese-language online course for OpenRefine, and I think it's very valuable to collect such language-specific resources somewhere, but this course was created in 2020 and perhaps there are more recent and better ones out there.

I would suggest that e.g. we perhaps use our hopefully-soon-to-be-launched Discourse forum to collect such (current, up to date) resources, and also encourage the community to keep them up to date (by language) there?

PR #67, opened 2020-03-21
Create aircraft

narcisse007:


wetneb:
Thank you for this aircraft! It's really kind of you :)

PR #68, opened 2020-04-09
Highlight solution for MacOS security problem

coopr:


thadguidry:
@coopr Thank you so much for this!  We're (right this very moment) applying to get a Apple Developer Program ID so that we can then begin signing and hopefully in a future release...make this problem go away.

PR #80, opened 2020-12-10
"Latter" -> "former"

claytonllibrar:
"Latter" means second of the two options, "former" would mean first, which I'm assuming is the intended meaning.


tfmorris (approved):

tfmorris:
Thanks for the contribution!

PR #82, opened 2021-01-15
Website link updates

allanaaa:
Changing most wiki links to the new docs.

There's a link on the Community page that I wanted to ask about before changing: 
https://github.com/OpenRefine/OpenRefine/blob/master/CONTRIBUTING.md
to 
https://docs.openrefine.org/technical-reference/contributing

### DIFF
--- a/documentation.md
+++ b/documentation.md
@@ -11,13 +11,13 @@ sidebar_link: true
 
   <h2 id="documentation-for-user">Documentation for users</h2>
 <ul>
-   <li><a href="https://github.com/OpenRefine/OpenRefine/wiki/">Official documentation</a></li>
+   <li><a href="https://docs.openrefine.org/">Official documentation</a></li>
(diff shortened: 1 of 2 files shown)


wetneb (approved):
Wohooo, this feels super good!

allanaaa:
Thoughts on the technical reference Contributing page?

wetneb:
Sorry I read too fast! Good question, the content is not quite the same. I guess we might need to keep a `CONTRIBUTING.md` file in the repository because it is sort of a standard location, but I would not be against migrating its contents to the Docusaurus docs and only leaving a pointer to that there. I would not maintain the same content in two places.

tfmorris:
We need the content on Github both so that it's part of the checkout for developers and because it's referenced by Github when it prompts new contributors to read the contribution guidelines.

https://docs.github.com/en/github/building-a-strong-community/setting-guidelines-for-repository-contributors

allanaaa:
Excellent. Assuming the contribution guidelines are important, I've added them to the Documentation page as well. I think this is good to go.

PR #86, opened 2021-02-26
Add reference to currently unavailable VIB-BITS extensions

ostephens:
Based on #74 by @pyrog this adds references to the (currently unavailable) VIB-BITS diff and save facets extensions (which can be found via Internet Archive's Wayback machine and maybe re-released for OpenRefine 3.4 (or upwards) at some point


thadguidry:
@ostephens ah nice, thanks for doing this.

allanaaa:
Two things: I tested diff in 3.3 and it works (no highlighting as per manual [is that even possible??] but the [- text] and [+ text] notations appeared as expected. So, I tested it in 3.4.1 and it works too! 
Save facets, alas, does not work.
The other thing is I would add (and did add) links directly to the manuals since those are up on Wayback too.

allanaaa (approved):

PR #88, opened 2021-07-27
adds C# .net core client to download page

ADelRosarioH:
Hi!

I recently developed a .NET Core client supporting the latest version of OpenRefine. You can take a look in here: https://github.com/ADelRosarioH/OpenRefine.Net.

I would like it to be in the client libraries section of https://openrefine.org/download.html and that's why I'm creating this PR.

I couldn't find any contributing guides for this repo, If there is one, I can close this PR and submit another according to the guide.

Thank you very much!


wetneb (approved):
Thanks for developing this library!

PR #95, opened 2021-11-30
Update blog post draft to correct Joey's name

trnstlntk:
I have no clue if I'm doing the right thing here.


wetneb:
Nice one! You definitely *can* do this, but it is quite unconventional :)
The more conventional way to do this would be to add a further commit directly on your `trnstlntk-OR-blogposts` branch, and push that branch to update the pull request.

By the way, your branch name suggests that you might have intended to use it for multiple blog posts (so, multiple PRs), and that should also generally be avoided: it is easier to create one branch per pull request.

wetneb:
No longer needed since the correction was included in the original PR

PR #101, opened 2022-02-22
update year to 2022

annajiat:
updated year from 2021 to 2022

Proposed
The latest stable release of OpenRefine 3.5, released on January 26, 2022. 

Current
The latest stable release of OpenRefine 3.5, released on January 26, 2021.


wetneb (approved):
Thanks for that!

PR #115, opened 2022-09-20
Rename 2022-09-20-discourse to 2022-09-20-discourse.md

thadguidry:
Fix file name to use markdown extension (which should fix index and not displaying issue)


trnstlntk (approved):
Thank you!

PR #119, opened 2022-10-06
Added MetricDoc Data Quality extension to list

thadguidry:


ostephens:
I tried to use this extension with 3.4.1 but got a butterfly error. Looking at the github repo, it looks like there could have been issues running the extension since 3.1 https://github.com/christianbors/OpenRefineQualityMetrics/issues/1

@thadguidry  how were you able to get this running with more recent versions of OR?

trnstlntk:
I currently don't have the time or brainspace to test/review this. Will gladly take Owen's word.

thadguidry:
@ostephens I used it back then, but not currently and why I suggested that it could use some love and perhaps a grant to convince Christian to update it.  I think we can just change the version compatible in my patch?

ostephens:
>I think we can just change the version compatible in my patch?
Agreed - but I don't currently have an installation older than 3.4.1 so can't check what the latest compatibility immediately

wetneb:
Since it does not seem to work on recent versions of OpenRefine, let's close this.

Issue #130, opened 2022-11-24
Fix old website links


trnstlntk:
With our shiny new website, old hyperlinks to former URLs need a check, and should redirect correctly.

I just followed a link to https://openrefine.org/blog/2022/06/28/2022-survey-results.html (which is linked in a slideshow I frequently use, and have shared with many people) and it made my browser very sad (there seems to be a loop of redirects or something). The URL ends up becoming https://openrefine.org/blog/2022/06/28/2022-survey-results.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html ;-)

<img width="974" alt="image" src="https://user-images.githubusercontent.com/9881645/203760204-fe695d1b-55af-4678-99eb-3e48c15c1a8d.png">

trnstlntk:
Links to images also seem to be broken, example https://openrefine.org/img/2022survey/2022-survey-communities.png linked from https://openrefine.org/blog/2022-survey-results

wetneb:
The redirects should work as expected now.
For the images, yes that seems related to the single-page black magic that Docusaurus uses. The solution to that would be to rewrite the blog posts to use proper markdown syntax, and not HTML tags.

PR #131, opened 2022-11-24
Fix image links

ostephens:
This should fix the image links in the blog posts as per #130
The images should ideally have alt tags added but I've left these out for now to prioritising the link fix


wetneb (approved):
Yay, it works! Thanks for taking care of this!

PR #163, opened 2022-12-05
Add documentation for cell rendering extension point

wetneb:
For https://github.com/OpenRefine/OpenRefine/pull/5312

### DIFF
--- a/versioned_docs/version-3.6/technical-reference/writing-extensions.md
+++ b/versioned_docs/version-3.6/technical-reference/writing-extensions.md
@@ -207,6 +207,44 @@ MenuSystem.appendTo(menu, ["core/facet"], [
 
 In addition to `MenuSystem.appendTo`, you can also call `MenuSystem.insertBefore` and `MenuSystem.insertAfter` which the same 3 arguments. To see what IDs you can use, see the function `DataTableColumnHeaderUI.prototype._createMenuForColumnHeader` in [/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js](http://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js).
 
+#### Cell renderers {#cell-renderers}
+
+From OpenRefine 3.7 on, extensions can also customize the way cells are rendered. This is done by registering a renderer, which is responsible for transforming the JSON representation of a cell into DOM elements rendering it:
+
+```js
+class MyCellRenderer {
+
+  constructor() {
+    super();
+    // some initialization code can be added here
+  }
+
+  render(
+    rowIndex, // the 0-based row index
+    cellIndex, // the position of the column to which the cell belongs
+    cell, // the deserialized JSON representation of the cell
+    cellUI // the parent CellUI object which called this renderer
+  ) {
+     // this renderer has the opportunity to return a DOM element represeting the cell, as follows:
+     return $('<span>rendered cell</span>');
+     // or it may not return anything, in which case the next cell renderer will be executed
+  }
+}
+```
+
+To render a cell, OpenRefine will execute each cell renderer in order, until the first renderer which returns a DOM element. The following renderers are not executed and that DOM element is used as the cell representation.
+
+The registration of the renderer is done with
+```js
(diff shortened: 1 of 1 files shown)


versioned_docs/version-3.6/technical-reference/writing-extensions.md
```diff
@@ -207,6 +207,44 @@ MenuSystem.appendTo(menu, ["core/facet"], [
...
+The registration of the renderer is done with
+```js
+CellRendererRegistry.addRenderer(
+    'my-renderer-identifier', // a string identifying our new renderer
+    new MyCellRenderer(), // the renderer itself
+    'recon' // the existing renderer it should be inserted before
+ );
+```
+
+The default renderers available in OpenRefine itself can be found in `main/webapp/modules/core/scripts/views/data-table/cell-renderers/registry.js`.
```
thadguidry, line 246: This also doesn't say where extension should **register** their cell renderers.  Maybe we could say something about that?  Is it in the main `../cell-renderers/registry.js` or do we have some other place for extensions to create a registry of their own that we then parse?

thadguidry (changes requested):

versioned_docs/version-3.6/technical-reference/writing-extensions.md
wetneb: > This does not say where extensions should register their cell renderers.

The `CellRendererRegistry.addRenderer` call is the registration - there is nothing else for an extension to do beyond making this call.

I do not see why extensions would want to create a registry of their own.

versioned_docs/version-3.6/technical-reference/writing-extensions.md
```diff
@@ -207,6 +207,44 @@ MenuSystem.appendTo(menu, ["core/facet"], [
...
+    cellUI // the parent CellUI object which called this renderer
+  ) {
+     // this renderer has the opportunity to return a DOM element represeting the cell, as follows:
+     return $('<span>rendered cell</span>');
+     // or it may not return anything, in which case the next cell renderer will be executed
+  }
+}
+```
+
+To render a cell, OpenRefine will execute each cell renderer in order, until the first renderer which returns a DOM element. The following renderers are not executed and that DOM element is used as the cell representation.
```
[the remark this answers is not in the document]
wetneb: I added a sentence before to make it clear that the renderers are held in the registry in a specific order.

versioned_docs/version-3.6/technical-reference/writing-extensions.md
thadguidry: Great!

versioned_docs/version-3.6/technical-reference/writing-extensions.md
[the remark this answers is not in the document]
thadguidry: Great, now I understand clearly. Thanks!

versioned_docs/version-3.6/technical-reference/writing-extensions.md
```diff
@@ -207,6 +207,44 @@ MenuSystem.appendTo(menu, ["core/facet"], [
 
 In addition to `MenuSystem.appendTo`, you can also call `MenuSystem.insertBefore` and `MenuSystem.insertAfter` which the same 3 arguments. To see what IDs you can use, see the function `DataTableColumnHeaderUI.prototype._createMenuForColumnHeader` in [/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js](http://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js).
 
+#### Cell renderers {#cell-renderers}
+
+From OpenRefine 3.7 on, extensions can also customize the way cells are rendered. This is done by registering a renderer, which is responsible for transforming the JSON representation of a cell into DOM elements rendering it:
```
thadguidry, line 212:
```suggestion
From OpenRefine 3.7 onwards, extensions can also customize the way cells are rendered. This is done by registering a renderer, which is responsible for transforming the JSON representation of a cell into DOM elements rendering it:
```

small suggestion to make translations later easier.

thadguidry (approved):

PR #166, opened 2022-12-11
Removel

Marcoscormo:


wetneb:
hi @Marcoscormo, it is not clear to me what you are proposing with this PR. Could you describe in a few words what you tried to do? It would be worth doing it again starting from the latest commit on the `master` branch.

PR #167, opened 2022-12-13
fix typo for pull request urls

thadguidry:


wetneb (approved):
Thanks, I should have checked that at least…

Issue #157, opened 2022-11-24
Provide a short intro for the blog posts


wetneb:
Docusaurus lets us [define a truncation mark](https://docusaurus.io/docs/blog#adding-posts) in blog posts by adding `<!--truncate-->` on an empty line. This defines the separation between the introduction of the blog post and the rest of its contents. This would be helpful to provide a more compact list of posts on https://openrefine.org/blog

We could go through the last few blog posts and try to figure out a sensible truncation location, so that the blog is a bit easier to skim through.

wetneb:
Closed by #169

PR #169, opened 2022-12-19
add <!--truncate--> for #157

magdmartin:
@wetneb is this what you are looking for to solve #157 ?
I like to confirm the format before going through every single post.


wetneb (approved):
Indeed!

PR #171, opened 2022-12-30
EOSS5 announcement blog post

wetneb:
Here is a draft blog post to announce the EOSS5 grant.

### DIFF
--- a/blog/2022-12-30-eoss5.md
+++ b/blog/2022-12-30-eoss5.md
@@ -1,6 +1,6 @@
 ---
 author: Antonin Delpeuch
-title: "OpenRefine funded again as an Essential Open Source Software for Science to improve its reproducibility"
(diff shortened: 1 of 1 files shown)


thadguidry:
@wetneb Welcome news!  I sincerely encourage you to be bold in re-imagining Change and History.  In particular, there is A LOT to say about a particular `HistoryEntry` and in fact, that's even a terrible class name for what is needed with an eye towards reproducibility in the future.  I would say **be bold** here and throughout with data structure changes as necessary.  I'd also encourage you to inspect commit comments from David Huynh through the years past all the way back where lots of insight was given about design directions and potential future changes never realized.

wetneb:
Thanks Thad! Yes, I am sure this will result in quite some changes in how the history is represented and serialized. I am also keen to try bold things out and am excited by the project.

PR #173, opened 2023-01-03
Wrong extensions link on documentation

robertgarrigos:
Fixes #172

PR #174, opened 2023-01-04
Remove docs versioning

### LINKED ISSUE #140: Documentation: version only by major version
We currently offer separate documentation for each minor release of the tool (3.4, 3.5, …).
This has a few downsides:
* when making improvements to the docs, we often need to edit multiple copies of the docs, which is cumbersome
* not much is changing between each minor version anyway

At the same time, we do not have a good mechanism to publish the documentation of the 4.0 branch.
If we create a `4.0` version in Docusaurus, it will be the one displayed by default to the users (because it is the latest one), but since we have not published a stable release in that branch, this is undesirable (people should see 3.x first).

### Proposed solution

Switch to only one documentation per major version instead. This means that at the moment, we would have version for OpenRefine 3, and the development docs would be for OpenRefine 4.

### Alternatives considered

I am not sure how else we could publish docs for 4.0 with the current system.

### Additional context

Also, that solution would mean that we would need to edit the docs for OpenRefine 4 on the master branch, which might be a bit confusing. If we also want to migrate our openrefine.org website to Docusaurus, perhaps this is a sign we should be using a different repository after all?

wetneb:
Closes #140.

In #140 I suggested to document by major version only (so, to have one copy of the docs for 3.x and one for 4.x) but since the docs for 4.x are not even here yet, I propose to just disable versioning entirely for now. This will avoid problems like in #173 where people edit the wrong files.


ostephens:
I'm slightly concerned this leaves us with no way of indicating what is present/absent from different versions of OpenRefine. I'm not against rationalising the number of versions we support in the documentation, but just removing the versioning completely doesn't feel right to me - at least without some other way of indicating which version some function or GREL is available for

wetneb:
> some other way of indicating which version some function or GREL is available for

That is indeed what I have tried to do, see for instance this change:
https://github.com/OpenRefine/openrefine.org/pull/174/files#diff-8fcfdd80888e73856ebcb046724748a401aff238be593c36ff72fa19515aecebR150

I went through all files and checked the difference between "next" (3.7), "3.6", "3.5" and "3.4", and tried to highlight those differences in the "next" version of the docs (which is the only one remaining after this PR).

ostephens:
I'm so sorry @wetneb  - I had a gut reaction to the point of the PR and commented without checking the detail :( Really sorry

wetneb:
I should have described the approach in more detail in the first place! Expecting people to eyeball changes to 125 files is not great ^^

PR #188, opened 2023-01-28
Improve architecture page, updating and filling gaps

wetneb:
Some parts of this content will need to be adapted for 4.x. Because we are not going to discontinue 3.x immediately I think it would be worth keeping both, and I am thinking of using the tab system (used for operating systems in other pages) to let the reader switch between the two, depending on which architecture they are working with.

Also, quite some comments about the roles of specific Java classes would be better done as proper Javadocs. In 4.x I have been working towards doing more of that, and I hope to be able to publish such javadocs for some core modules at least (such as `refine-model`).

### DIFF
--- a/docs/technical-reference/architecture.md
+++ b/docs/technical-reference/architecture.md
@@ -18,7 +18,7 @@ server- and client-side. A [list of known extensions](https://openrefine.org/ext
 The client-side part of OpenRefine is implemented in HTML, CSS and plain Javascript. It primariy uses the following libraries:
 * [jQuery](http://jquery.com/)
 * [Wikimedia's jQuery.i18n](https://github.com/wikimedia/jquery.i18n)
-The front-end dependencies are fetched at build time via NPM.
+The front-end dependencies are fetched at build time via [NPM](https://www.npmjs.com/).
 
 The server-side part of OpenRefine relies on many libraries, for instance to implement import and export in many different formats.
 Those are fetched at build time via [Apache Maven](https://maven.apache.org/).
@@ -36,7 +36,7 @@ As mentioned before, the server-side maintains states of the data, and the prima
 
 ### Projects {#projects}
 
-In OpenRefine there's the concept of a workspace similar to that in Eclipse. When you run OpenRefine it manages projects within a single workspace, and the workspace is embodied in a file directory with sub-directories. The default workspace directories are listed [in the manual](manual/installing.md#set-where-data-is-stored) and it also explains how to change them.
+In OpenRefine there's the concept of a workspace similar to that in [Eclipse IDE](https://www.eclipse.org/ide). When you run OpenRefine it manages projects within a single workspace, and the workspace is organized in a file directory with sub-directories. The default workspace directories are listed [in the manual](manual/installing.md#set-where-data-is-stored) and it also explains how to change them.
 
 The class `ProjectManager` is what manages the workspace. It keeps in memory the metadata of every project (in the class `ProjectMetadata`). This metadata includes the project's name and last modified date, and any other information necessary to present and let the user interact with the project as a whole. Only when the user decides to look at the project's data would `ProjectManager` load the project's actual data. The separation of project metadata and data is to minimize the amount of stuff loaded into memory.
 
@@ -204,7 +204,7 @@ See `main/webapp/modules/core/scripts/index/default-importing-sources/sources.js
 
 #### File Selection Panel {#file-selection-panel}
 
-This screen is shown when there are multiple files to choose from when creating a project, for instance after uploading a zip file with multiple files in it. This interface lets the user choose which ones to import.
+This screen is shown when there are multiple files to choose from when creating a project, for instance after uploading a zip file with multiple files in it. This interface lets the user choose which files to import to create a new project.
 Although OpenRefine only supports one table per project so far, it is possible to select multiple files to import. Their contents will be concatenated into a single table.
 
 #### Parsing UI Panel {#parsing-ui-panel}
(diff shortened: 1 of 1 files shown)


docs/technical-reference/architecture.md
```diff
@@ -204,25 +203,61 @@ The argument `form` is a jQuery-wrapped FORM element that will get submitted to
...
 
-#### UrlRewriter {#urlrewriter}
-Documentation not currently available
+An importing controller is a component of the back-end which is in charge of the entire importing workflow, from the initial transfer of the raw data to be imported to the created project, with all the configuration steps in between, [as described in the
+earlier section](#importing-controllers). OpenRefine comes with
+a default importing controller which implements this for data coming from:
+* file upload by the user via the web interface
+* upload of textual information using the clipboard import form
+* download of a file by supplying a URL
+
```
thadguidry, line 236: :::note
The importing controller is responsible for *new project* creation and is not responsible for opening or importing *existing projects*, which is the responsibility of the `ProjectManager` class.
:::

docs/technical-reference/architecture.md
```diff
@@ -204,25 +203,61 @@ The argument `form` is a jQuery-wrapped FORM element that will get submitted to
...
+For all of these data sources, the first step consists storing the corresponding input files in a temporary directory inside the workspace. The default importing controller provides an HTTP API used by the front-end to select which files to import, [predict
```
thadguidry, line 237:
```suggestion
For all of these data sources, the first step consists of storing the corresponding input files in a temporary directory inside the workspace. The default importing controller provides an HTTP API used by the front-end to select which files to import, [predict
```

docs/technical-reference/architecture.md
```diff
@@ -204,25 +203,61 @@ The argument `form` is a jQuery-wrapped FORM element that will get submitted to
...
+the format they are in](#formatguesser), provide default importing options for the selected format, preview the project's first few rows with the given options, and finally create the project.
+
+Extensions can define other importing controllers to implement other importing flows depending on the data source. For instance, importing data from a SQL database requires different steps such as selecting the database and providing a SQL query. The
+`database` extension implements such a workflow by providing its own importing controller.
 
 #### FormatGuesser {#formatguesser}
-Documentation not currently available
+
+A format guesser is a class that tries to determine the MIME type of a file, considering its contents.
```
thadguidry, line 245:
```suggestion
A format guesser is a class that tries to determine the MIME type of a file, considering its contents and/or file extension.
```
I think the above is more correct now since @tfmorris did changes on a few in the past?

docs/technical-reference/architecture.md
```diff
@@ -204,25 +203,61 @@ The argument `form` is a jQuery-wrapped FORM element that will get submitted to
...
+This could be avoided by trying to read the given files with the predicted importer before suggesting the format to the user, making sure that at least that does not throw an exception.
 
 #### ImportingParser {#importingparser}
-Documentation not currently available
 
+An `ImportingParser` is a class that is responsible for parsing a file into OpenRefine's project model.
+It takes a range of importing options passed on from the frontend, input by the user into a dedicated UI, specific to the format being parsed.
+
+When possible, parsers are designed so that they can import the first few rows of the project without reading the entire input file in memory. This helps provide fast previews of the project to be created when the user changes importing options. Every
+change in the importing options triggers a new parse of the source files.
```
thadguidry, line 260:
```suggestion
change in the importing options triggers a new parse of the source files (unless the user has deselected that option in the parsing UI panel).
```

thadguidry (changes requested):

docs/technical-reference/architecture.md
thadguidry: This is one area that often has confused a few contributors in the past.  So maybe this note admonition can help clear things up a bit.

docs/technical-reference/architecture.md
thadguidry: grammar improvement

docs/technical-reference/architecture.md
```diff
@@ -37,13 +36,13 @@ As mentioned before, the server-side maintains states of the data, and the prima
...
+In OpenRefine there's the concept of a workspace similar to that in Eclipse. When you run OpenRefine it manages projects within a single workspace, and the workspace is embodied in a file directory with sub-directories. The default workspace directories are listed [in the manual](manual/installing.md#set-where-data-is-stored) and it also explains how to change them.
 
 The class `ProjectManager` is what manages the workspace. It keeps in memory the metadata of every project (in the class `ProjectMetadata`). This metadata includes the project's name and last modified date, and any other information necessary to present and let the user interact with the project as a whole. Only when the user decides to look at the project's data would `ProjectManager` load the project's actual data. The separation of project metadata and data is to minimize the amount of stuff loaded into memory.
```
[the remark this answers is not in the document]
wetneb: I think it's already clear like that because of the last sentence. Yes if something is in memory, it is by definition already loaded.

docs/technical-reference/architecture.md
```diff
@@ -37,13 +36,13 @@ As mentioned before, the server-side maintains states of the data, and the prima
...
 
 A project's _actual_ data includes the columns, rows, cells, reconciliation records, and history entries.
 
-A project is loaded into memory when it needs to be displayed or modified, and it remains in memory until 1 hour after the last time it gets modified. Periodically the project manager tries to save modified projects, and it saves as many modified projects as possible within 30 seconds.
+A project is loaded when it needs to be displayed or modified, and it remains in memory until 1 hour after the last time it gets modified. Periodically the project manager tries to save modified projects, and it saves as many modified projects as possible within 30 seconds.
```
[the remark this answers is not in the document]
wetneb: I'd say this is redundant as well, I'd rather keep it concise.

docs/technical-reference/architecture.md
wetneb: Technically speaking the FormatGuesser is only looking at the file contents: the file extension is taken into account elsewhere.

thadguidry: Ah, ok.

docs/technical-reference/architecture.md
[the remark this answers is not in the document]
thadguidry: Sure.

docs/technical-reference/architecture.md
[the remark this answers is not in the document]
thadguidry: Fine, but I think it's more clear to mention the disk to loading in memory in the same sentence.

PR #194, opened 2023-02-13
Add page describing architecture of version 4

wetneb:
Draft of docs describing the new architecture of version 4.
I plan to improve this in the future as there are still aspects of this architecture which will need improving over the course of the reproductibility project.


wetneb:
Direct link to the interesting part: https://deploy-preview-194--openrefine-website.netlify.app/docs/technical-reference/architecture-4

thadguidry:
Awesome!  Should we wait on applying various grammar fixes? or wait a month or so?  Even some of the Javadocs read slightly ambiguous, like they are missing commas or something?  This seems longwinded or missing commas?

>Change.ChangeResult
>Bundles up various pieces of information: the new grid after applying the change whether the rows or records of the original grid were preserved a representation of the dependencies of this change

wetneb:
For the docusaurus part, we can very well do copy-editing now already.
For Javadocs, it is likely that it is a formatting issue due to Javadoc's own rendering, so PRs on the 4.0 branch would be needed for that. If you feel like having a go at that it is of course most welcome!

PR #197, opened 2023-03-01
Updated Maven documentation

elebitzero:
Updated Maven documentation for https://github.com/OpenRefine/OpenRefine/issues/5484.


wetneb (commented):
Perhaps it's useful to keep the examples to make it clear that the `bin` subdirectory should not be included in the path?

wetneb (approved):

Issue #122, opened 2022-10-26
Link to blog RSS feed


tfmorris:
I don't know if anyone uses RSS feed readers any more, but since the blog has an RSS feed, we might as well publicize it by providing a link https://openrefine.org/feed.xml someplace (or places).

thadguidry:
@tfmorris where would you publicize it?  Perhaps a link next to the title? or as last item in menu?

![image](https://user-images.githubusercontent.com/986438/198173149-d877254a-75f5-42ab-b5a8-5d4fec4f210a.png)

tfmorris:
Either/both is fine. If you use the RSS icon, it can go next to the Blog menu item rather than needing a separate line.

amyra98:
So I need to add RSS Feed text beside Blog and when clicked it will go to https://openrefine.org/feed.xml ?
Can you tell me the location or send the link for the content shown in screenshot?
Please assign me this issue
@tfmorris @thadguidry

wetneb:
The page which corresponds to this old screenshot is now here:
https://openrefine.org/blog

PR #211, opened 2023-04-01
RSS Feed feature

amyra98:
The changes are done corresponding to #122 
Since the blog has an RSS feed, a link and RSS Feed have been added to the navbar.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -39,6 +39,10 @@ module.exports = {
           position: 'left',
         },
         {
+          label: 'RSS Feed',
+          href: 'https://openrefine.org/feed.xml',
+        },
+        {
           to: '/donate',
           label: 'Donate',
           position: 'right',
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -22,6 +22,7 @@ module.exports = {
           label: 'Download',
           position: 'left',
         },
+        
         {
           to: '/docs',
           activeBasePath: 'docs',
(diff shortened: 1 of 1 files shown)


wetneb:
Thanks!
You have added the link on the same level as "Download", "Documentation", "Community", "Blog". Intuitively the RSS link should be internal to the Blog category (so that it is clear that it's the RSS feed of blog entries and not, say, of OpenRefine releases…). That might be a little more difficult to insert than in this location, but still perhaps worth the effort, no?

amyra98:
Yes, I'll work on it

amyra98:
Please review @wetneb

wetneb:
Thanks! I was rather thinking that the link would be visible inside the blog page itself, not in the menu.
With the approach you have taken, the RSS feed is a little bit "in the way" in the sense that people who will want to visit the blog will first need to open the "Blog" menu and then click on the "Blog" entry in it, it does not feel very fluid to me.

How about something like this? Potentially with an additional RSS icon before the link itself:
image

I think you could do this by changing the configuration of the blog plugin in docusaurus to use an alternate component for the blog main page: 
https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-blog#blogListComponent
It's not as easy as changing a configuration file for sure, but I would say it is worth it for usability.
Also, we could consider upstream this to Docusaurus itself, because there does not seem to be any out-of-the-box option to display an RSS link in the blog, and we are likely not the only website which would like to have that.

thadguidry:
There's also some really good discussion if you [search "rss" across their discussions](https://github.com/facebook/docusaurus/search?q=rss&type=discussions) like this [one](https://github.com/facebook/docusaurus/discussions/5623)

Atom & RSS are included and turned on by default in Docusaurus configurations.  The links typically show along the bottom in the footer of a blog configured site.  Such as this one: https://johnnyreilly.com/  John happens to be a valued contributor to Docusaurus and it might be good to check-in with him also about this issue and see what he says or thinks.  Good Luck!

PR #218, opened 2023-04-26
Fix cross function example

wetneb:
It looks like the only example of the `cross` function we had in the docs was actually wrong… Can anyone confirm?

### DIFF
--- a/docs/manual/grelfunctions.md
+++ b/docs/manual/grelfunctions.md
@@ -601,7 +601,7 @@ Returns the first non-null from a series of objects. For example, `coalesce(valu
 ###### cross(cell, s projectName (optional), s columnName (optional)) {#crosscell-s-projectname-optional-s-columnname-optional}
 Returns an array of zero or more rows in the project projectName for which the cells in their column columnName have the same content as the cell in your chosen column. For example, if two projects contained matching names, and you wanted to pull addresses for people by their names from a project called “People” you would apply the following expression to your column of names: 
 ```
-cell.cross("People","Name").cells["Address"].value[0]
+cell.cross("People","Name")[0].cells["Address"].value
 ```
 
 This would match your current column to the “Name” column in “People” and, using those matches, pull the respective “Address” value into your current project. 


thadguidry:
Yes, the example is wrong in the sense that an Array might be the result, or it might not.  Depends on if it returns an `[ [ object Row ] ]` Array technically?  For now that's not likely to change? Until we introduce storing Arrays in cells?

The original expression syntax was to get the 1st Row from the Array 
`cell.cross("TestCross","Name")[0]` et.al. @dfhuynh 
But dunno if there is always 1 Row or Record only now after @antoine2711 's changes...which is slightly concerning, but we'll figure it out.

image
image
image
image

antoine2711:
@wetneb : yes, the documentation is wrong.

`cell.cross("People","Name").cells["Address"].value[0]`

should of course be

`cell.cross("People","Name")[0].cells["Address"].value`

Regards, A.

antoine2711:
> But dunno if there is always 1 Row or Record only now after @antoine2711 's changes... which is slightly concerning, but we'll figure it out.

No, this has not changed with the fix 3 years ago. Somehow, the doc got broken. Sad, but good thing it's now fixed.

Regards,
   Antoine

ostephens:
In fact both expressions work. Although `cross` (as it has always done afaik) returns an array of rows, it seems that `.cells[]` will iterate of the array of rows and and `value` will iterate over the array of rows. So:

`cell.cross("People","Name")` -> array of row objects
`cell.cross("People","Name").cells["Address"]` -> array of cell objects
`cell.cross("People","Name").cells["Address"].value` -> array of values

So 
`cell.cross("People","Name").cells["Address"].value[0]`
and
`cell.cross("People","Name").cells["Address"][0].value`
and
`cell.cross("People","Name")[0].cells["Address"].value`

all give the same result

I'm not sure if this has always been the case, but I always use (and teach) the last of these as it makes most sense to me, but I guess the other options provide some convenience if you just need a single value

wetneb:
Ah, good point. Actually, I find it really messy that GREL silently applies a computation to the elements of an array as a map.

tfmorris:
All of these examples deemphasize the fact that `cross()` can return multiple rows. I think it would be better for the example to using `join()` to highlight handling of multiple values.

PR #224, opened 2023-06-14
Document minimum version of dev dependencies

wetneb:
Follow-up for https://github.com/OpenRefine/OpenRefine/pull/5916


elebitzero:
@wetneb, we can enforce the Node version using the engines field in the package.json. I'll submit a PR for that.

PR #233, opened 2023-07-27
Split contributing page into role-specific subpages

### LINKED ISSUE #232: Re-structure the guide for new contributors by splitting it into separate pages
The [Contributing](https://openrefine.org/docs/technical-reference/contributing) page in our documentation serves as entry point for new contributors. It has grown a bit wild as we have added information to it, in a rather ad-hoc way.

We should split it in different pages for each type of contributor, so that prospective contributors do not get lost in an ocean of instructions, most of which are irrelevant to them. I would make a page for:
* developers
* designers
* translators
* documentation contributors
* maybe also for user support and issue reporting?

Beyond splitting the existing content into those separate pages, I think we can improve the content itself, but we can do that in a second step.

This is a follow-up on a discussion with @magdmartin

wetneb:
Closes #232.
As mentioned there, I think the content itself can be improved too (which I would do as a follow-up PR).

See the new "contributing" page here:
https://deploy-preview-233--openrefine-website.netlify.app/docs/technical-reference/contributing


magdmartin (approved):

magdmartin:
@wetneb  if you are ok with the two links I added, then it is good to merge

PR #242, opened 2023-09-07
Expose release information in /versions.json

### LINKED ISSUE #241: Expose version information in a machine-readable way
OpenRefine currently queries the GitHub API to check for new updates. Instead, it should download a file from this website, which we control more than GitHub.

See:
* https://github.com/OpenRefine/OpenRefine/issues/6017
* https://github.com/OpenRefine/OpenRefine/issues/4463

wetneb:
Closes #241.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -142,7 +142,7 @@ module.exports = {
     },
   },
   themes: [],
-  plugins: [],
+  plugins: ['./src/plugins/docusaurus-versionsjson-plugin'],
   presets: [
     [
       '@docusaurus/preset-classic',
--- a/releases.json
+++ b/releases.json
@@ -1,0 +1,75 @@
+[
+    {
+        "version": "3.7.5",
+        "stable": true,
+        "date": "2023-09-11",
+        "source": "github",
+        "artifacts": [
+             {
+                "platform": "win-with-java",
+                "format": "zip"
+             },
+             {
+                "platform": "win",
+                "format": "zip"
+             },
+             {
+                "platform": "mac",
+                "format": "dmg"
+             },
+             {
+                "platform": "linux",
+                "format": "tar.gz"
+             }
+        ]
+    },
+    {
+        "version": "3.6.2",
+        "stable": true,
+        "date": "2022-10-03",
+        "source": "sonatype",
+        "artifacts": [
+             {
+                "platform": "win-with-java",
+                "format": "zip"
+             },
+             {
+                "platform": "win",
+                "format": "zip"
+             },
+             {
+                "platform": "mac",
+                "format": "dmg"
+             },
+             {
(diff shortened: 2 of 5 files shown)


wetneb:
See output at https://deploy-preview-242--openrefine-website.netlify.app/versions.json

tfmorris:
Is there a reason to not just use URLs for the downloads instead of publishing pieces that need to be concatenated along with some external information?

wetneb:
We can add the URLs for sure, it takes a bit more effort but it should be easy too. This PR is a sort of minimal change which intends to unblock the changes on OpenRefine's side (it took me quite some time to figure out how to integrate it with Docusaurus' build process).

For now I have just reused the format of the JSON blob that underpins the generation of the download page, which is the part that we edit manually when making a release. That's why URLs are not directly present in that (because making a release would involve tweaking the version number in multiple URLs, which is error prone).

elebitzero (commented):
Why do we need to publish versions.json when we already have a releases.json?
version.json just adds a block ` releases: <releases.json> }`

It seems easier to publish the original release.json.

wetneb:
I see those as serving two slightly different purposes:
* `releases.json` is a file that is easy to edit by hand for a maintainer doing a release. If I publish a `3.7.5` release, all I need is to edit a single character in that file (change `4` to `5`). The format of this file can evolve as our needs to generate the website change - maybe one day we even delete it entirely and call the GitHub API when the site is generated, or things like that
* `versions.json` is a file that is meant to be used as an API. For instance, it would make sense to include the download URLs of each artifact as suggested by @tfmorris. We should ensure that the format is mostly stable (apart from adding new fields, or other compatible changes) since old OpenRefine versions will be relying on it.

Also the reason why I introduced a root object in the format of versions.json is that having a root object makes it easier to add more information in a compatible way later on (we can add new keys in it, which is not possible if you have a root array).

wetneb:
To demonstrate why it can be useful to have `versions.json` and `releases.json` differ I added the URLs Tom requested:
https://deploy-preview-242--openrefine-website.netlify.app/versions.json

tfmorris:
I forgot that we don't actually do any downloads, but instead rely on the user to do that, so the only URL that we really need is that of the download page. Having a hosted download page also allows us to remind them another couple of times that they should do backups before upgrading. 

fyi - Here's the original format: 
https://github.com/OpenRefine/OpenRefine/blob/bf60360b26c76f39d41107654dfb4549813b79a6/main/webapp/modules/core/scripts/index.js#L104-L118

It had a `releases` key and a `homepage` key with the download page at the top level. Since this is a once-a-decade migration, we should make sure that we're happy with the new format as we change locations.

wetneb:
Merging soon unless there are concerns.

elebitzero (approved):

PR #243, opened 2023-09-07
Improve page about documentation contributions

wetneb:
The current page is out of date (some bits are left from the situation when we had our docs in the main repository).
I tried to update it so that it is also more accessible to folks who are not familiar with the terminal or Git.

### DIFF
--- a/docs/technical-reference/documentation-contributions.md
+++ b/docs/technical-reference/documentation-contributions.md
@@ -24,7 +24,10 @@ npm install -g yarn
 ```
 after having installed Node.js.
 
-Once you have installed those tools, clone the [OpenRefine/openrefine.org](https://github.com/OpenRefine/openrefine.org) repository, open a terminal in your clone and run:
+Once you have installed those tools, create a private fork of the [OpenRefine/openrefine.org](https://github.com/OpenRefine/openrefine.org) repository and clone your fork, then change to the clone:
+
+git clone git@github.com:<myaccount>/openrefine.org.git
+cd openrefine.org
 ```sh
 yarn
 ```
--- a/docs/technical-reference/documentation-contributions.md
+++ b/docs/technical-reference/documentation-contributions.md
@@ -24,11 +24,11 @@ npm install -g yarn
 ```
 after having installed Node.js.
 
-Once you have installed those tools, create a private fork of the [OpenRefine/openrefine.org](https://github.com/OpenRefine/openrefine.org) repository and clone your fork, then change to the clone:
+Once you have installed those tools, create a personal fork of the [OpenRefine/openrefine.org](https://github.com/OpenRefine/openrefine.org) repository and clone your fork, then change to the clone:
 
+```sh
 git clone git@github.com:<myaccount>/openrefine.org.git
 cd openrefine.org
-```sh
 yarn
 ```
 This will install the dependencies required to generate the site, which is mainly [Docusaurus](https://docusaurus.io/).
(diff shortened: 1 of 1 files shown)


docs/technical-reference/documentation-contributions.md
```diff
@@ -4,34 +4,43 @@ title: Documentation contributions
...
-and will auto-refresh when you edit the source files:
+and will auto-refresh when you edit the source files.
+
+You could also generate the website as a set of static files, although this should generally not be needed:
 ```sh
-yarn start
+yarn build
 ```
 
+Once you are happy with your changes, you can submit them as a pull request.
```
tfmorris, line 44: Perhaps link this to a "your first pull request" page, since, if they are following these directions, they may not know some of the basics.

tfmorris (approved):
One suggestion, but otherwise looks good.

docs/technical-reference/documentation-contributions.md
wetneb: Do you mean https://openrefine.org/docs/technical-reference/code-contributions#your-first-code-pull-request ?
This one is meant to be specific for code PRs - at least I classified it in that page when splitting out the contributing guide.
Maybe we could generalize it but to be honest it feels like that part is not really OpenRefine-specific so I'd be happy linking to documentation about GitHub itself. But pages like [this](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests) are probably way too complicated and overwhelming. Maybe some third-party tutorial?

docs/technical-reference/documentation-contributions.md
```diff
@@ -4,34 +4,43 @@ title: Documentation contributions
 sidebar_label: Documentation contributions
 ---
 
-We use [Docusaurus](https://docusaurus.io/) for our docs. The repository for those docs is [OpenRefine/openrefine.org](https://github.com/OpenRefine/openrefine.org). For small documentation changes, you should be able to edit the Markdown files directly and submit them as a pull request. Each documentation page has a link at the bottom to help you do this in a few clicks. Once you submit your pull request, a preview of the docs will be generated automatically. 
+We welcome contributions to our documentation, which is maintained as a set of [Markdown](https://www.markdownguide.org/) files in the [OpenRefine/openrefine.org](https://github.com/OpenRefine/openrefine.org) Git repository.
```
magdmartin, line 8: Should we divide the page into two sections
1. Single page edit via GitHub web interface 
2. Multi-page edit by editing the documentation locally

wetneb: Isn't it already the case?

magdmartin: Sorry if my comment was not clear. My suggestion is to add a new h2 title "single page edit" and rename the other h2.

wetneb: Ok, I see! Perhaps it's simple enough if you edit the PR directly as you see fit?

docs/technical-reference/documentation-contributions.md
wetneb: ok that makes sense! I'd suggest you just go ahead and edit the PR as you see fit, I'm happy with all your changes

tfmorris: I'm about to leave for a race weekend, but I'll try to remember to do it Sunday night when I return.

PR #248, opened 2023-10-09
Update version-release-process.md

wetneb:
Simplify the list of steps to reflect recent changes in the release process:
* the artifacts are hosted on GitHub, not on Sonatype
* updating the website can be done by editing `releases.json`
* the docs no longer have separate snapshots for each version, removing one step in the release process
* Apple signing is done in the GitHub Actions - manual set up is not required

### DIFF
--- a/docs/technical-reference/version-release-process.md
+++ b/docs/technical-reference/version-release-process.md
@@ -19,10 +19,7 @@ git tag -a -m "Second beta" 2.6-beta.2
 7. [Update the OpenRefine Homebrew cask](https://github.com/OpenRefine/OpenRefine/wiki/Maintaining-OpenRefine's-Homebrew-Cask) or coordinate an update via the [developer forum](https://forum.openrefine.org/c/dev/8)
 8. Announce on the [announcements section of the OpenRefine forum](https://forum.openrefine.org/c/news/13), or even on the [blog](https://openrefine.org/blog) if this is a major release (the blog is imported automatically into the announcements category
     of the forum)
-9. Update the version in master to the next version number with `-SNAPSHOT` (such as `4.3-SNAPSHOT`)
-```shell
-mvn versions:set -DnewVersion=4.3-SNAPSHOT
-```
+9. Update the version in master to the next version number with `-SNAPSHOT` (such as `4.3-SNAPSHOT`), in the same places as in step 2.
 
 Apple code signing
 ==================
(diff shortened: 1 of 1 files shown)


docs/technical-reference/version-release-process.md
```diff
@@ -14,27 +14,20 @@ When releasing a new version of Refine, the following steps should be followed:
...
 ```shell
 mvn versions:set -DnewVersion=4.3-SNAPSHOT
 ```
-12. If releasing a new major or minor version, create a snapshot of the docs, following [Docusaurus' versioning procedure](https://docusaurus.io/docs/versioning).
 
 Apple code signing
 ==================
 
-We have code signing certificates for our iOS distributions. To use them, follow these steps:
+We have code signing certificates for our iOS distributions. Those are available in our build environment and you should not need to import them to your own machine to make a release (since you are not building the release yourself). But if the signing process, it can be useful to import the certificates on your own machine to debug the process. To do so:
```
magdmartin, line 30: I think the sentence is missing one word. Should it be `if the signing process fails` ?

docs/technical-reference/version-release-process.md
```diff
@@ -14,27 +14,20 @@ When releasing a new version of Refine, the following steps should be followed:
...
+6. Once the artifacts are ready, update the [`releases.json`](https://github.com/OpenRefine/openrefine.org/blob/master/releases.json) file in the openrefine.org repository to reflect this new version. Verify that the correct versions are shown at [http://openrefine.org/download](http://openrefine.org/download)
+7. [Update the OpenRefine Homebrew cask](https://github.com/OpenRefine/OpenRefine/wiki/Maintaining-OpenRefine's-Homebrew-Cask) or coordinate an update via the [developer forum](https://forum.openrefine.org/c/dev/8)
+8. Announce on the [announcements section of the OpenRefine forum](https://forum.openrefine.org/c/news/13), or even on the [blog](https://openrefine.org/blog) if this is a major release (the blog is imported automatically into the announcements category
     of the forum)
-11. Update the version in master to the next version number with `-SNAPSHOT` (such as `4.3-SNAPSHOT`)
+9. Update the version in master to the next version number with `-SNAPSHOT` (such as `4.3-SNAPSHOT`)
```
thadguidry, line 22: Where, how? If via a maven command what changes and does a PR push need to accompany it?

thadguidry (changes requested):

docs/technical-reference/version-release-process.md
wetneb: is that clearer like that?

PR #254, opened 2023-10-21
Put technical reference inside developer docs

wetneb:
I think the pages in the current "technical reference" category should rather go to the contributing instructions for developers.
This makes it clearer that this content is intended for developers and removes a top-level category, which is quite visible and prominent for users.

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -32,41 +32,10 @@ module.exports = {
     ],
    'GREL Reference': [
 	   'manual/grelfunctions'],
-    'Technical Reference': [
-      {
-        type: 'category',
-        label: 'Architecture',
-        items: [
-            'technical-reference/architecture-before-4',
-            'technical-reference/architecture-4',
-            'technical-reference/clustering-in-depth',
-            'technical-reference/openrefine-api',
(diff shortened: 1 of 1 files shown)


sidebars.js
```diff
@@ -32,41 +32,10 @@ module.exports = {
...
-         'technical-reference/version-release-process',
-         'technical-reference/homebrew-cask-process'
-        ]
-      }      
-  ],
   'Contributing to OpenRefine': [
     'technical-reference/contributing',
     'technical-reference/why-get-involved',
-
+    'technical-reference/github',
```
wetneb, line 38: I have also added this page back at this level because I think it's relevant for all types of contributors who get to interact on GitHub.

sidebars.js
```diff
@@ -95,6 +92,7 @@ module.exports = {
       },
 
       'technical-reference/documentation-contributions',
+      'technical-reference/development-roadmap',
```
wetneb, line 95: I have moved out the roadmap page from the "code contributions" section since it is relevant for all contributors

magdmartin:
It looks good to me. thanks for taking the time to make those changes.

PR #257, opened 2023-10-21
Document import order settings in IDE setup instructions

wetneb:
Follow-up to https://github.com/OpenRefine/OpenRefine/pull/6108


thadguidry (commented):
On L311 the "note" should probably be an admonition.  But it reads weirdly to me.  On L244 the file was mentioned. Import ordering is in header starting in L242 right?  Confusing.  Maybe a hyperlink to a sub heading for "import ordering" is better overall. Dunno.

PR #260, opened 2023-11-01
Add two extensions developed by RefinePro

wetneb:
I wasn't aware of them but @magdmartin pointed me to them in a call. I think they are worth listing even though they are targeting a fairly old version.


tfmorris:
Should the attribution be to the RefinePro forks or the original FAIRplus repos (e.g. https://github.com/FAIRplus/OpenRefine_Authenticator)? They appear to have been contributed by Novartis (NIBR).

wetneb:
> It is more likely that any new tickets or pull requests will receive attention on the RefinePro repositories.

Personally, I would say that the most actively maintained GitHub repository of a project should not be marked as a fork of another one, but as an independent one. You can convert a fork into a non-fork repository by making a request to GitHub's support.

tfmorris (changes requested):
I think it would be polite to acknowledge the funder & donor, unless they prefer to remain anonymous, but regardless, I strongly object to these extensions being promoted until their license violations are corrected. They contain entire modules of OpenRefine code, copied verbatim, without any of the required license notices.

For example the only differences in  [HttpClient.java](https://github.com/FAIRplus/OpenRefine_Authenticator/blob/main/src/main/java/com/refinepro/importing/HttpClient.java) are changing the package name and adding a single line (although it's also missing all the bug fixes we've made since it was copied).

wetneb:
It's a valid point, but I would assume good faith here and avoid to scare away the extension author from interacting with the OpenRefine project. I am glad they wrote an extension for OpenRefine and want to encourage them to do more of that.

Looking at the `HttpClient.java` file, they did make its provenance clear with a comment, so they probably thought they were correctly crediting the original authors by doing that:
```
/*
    Custom code from original OpenRefine project main/src/com/google/refine/util/HttpClient.java
*/
```
So I would suspect it's just a lack of familiarity with the concept of license header and the importance to retain them. I think it should be easy to fix.

PR #262, opened 2023-11-13
New funding page

magdmartin:
Introducing a new page listing all our funding (current, past and under application). 
This page will replace the current wiki page https://github.com/OpenRefine/OpenRefine/wiki/Funded-projects 
I'd like to move it to the website for 
* better visibility
* enabling source control for the different updates.


wetneb (commented):
Very nice, thanks a lot for putting it together! I made a quick pass over it to fix a few typos and changing the case away from "Title Case" to "Normal case".

For each, I think it would be nice to have a quick summary of the focus area of the grant, to get a sense of what sort of work gets funded (or not!). Would you be happy with that? I could try summarizing that if you want.

magdmartin:
Thank you for reviewing the document. I would like to propose that we publish it as it is, to make it readily available and so we can start referencing it in our communications. The blog post that is linked to each grant provides a good overview of what the grant is about. We can always add a summary for each grant later if we need to provide more detail.

PR #267, opened 2023-12-06
Update reconciling.md to include how to access Commons sitelinks (and labels, aliases, descriptions)

loraxian:
I think it is useful to include the way to access labels, aliases etc as people often struggle with this. Also I cannot find documented anywhere the way to get Commons sitelinks (adding the wiki on the end was not obvious to me!) so wanted to make sure that was added.


wetneb (approved):
Thank you!

PR #271, opened 2023-12-16
Add blog post to ask for BarCamp venues

wetneb:
I think it's worth making a blog post to reach out to folks who would be able to see this but don't watch the forum actively.

### DIFF
--- a/blog/2023-12-16-barcamp-venue-options.md
+++ b/blog/2023-12-16-barcamp-venue-options.md
@@ -3,7 +3,7 @@ author: Antonin Delpeuch
 title: Date and venue options for an OpenRefine team meeting in 2024
 ---
 
-For some time, a few team members have been considering organizing a small in-person meeting around OpenRefine.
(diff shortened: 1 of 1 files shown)


wetneb:
See preview here: https://deploy-preview-271--openrefine-website.netlify.app/blog/2023/12/16/barcamp-venue-options

Ainali:
I like it! But it is a bit ambiguous in first stating a small event, and then in the end have an open invitation for anyone. As those seem to be a bit contradictory. Perhaps instead of saying "a small in-person event", perhaps say "an in-person event for about X people" (substitute X for what we want to aim for)? That way, the ambitions are clear, and it would be easier for people to suggest venues that could fit.

wetneb:
Good point! The intention with stating that it's open to anyone is to make the event more inclusive in terms of roles (to avoid it being developer-centric), rather than asserting that we have an unbounded hosting capacity. I'll reformulate.

Ainali (approved):

Issue #263, opened 2023-11-17
Add back a top level ` docs/technical-reference/architecture` page to prevent broken links


ostephens:
Following the addition of the v4 architecture documentation, the original architecture documentation was renamed to `architecture-before-4` which means that links to the old `architecture` URL give a 404

I propose adding back a page at ` docs/technical-reference/architecture` which simply points to each of the new documents. The page does not need to be part of the sidebar navigation, but just to prevent 404 from existing links to the OpenRefine docs

aakash-a-dev:
Hello I raised a Pull Request on this issue. Will like you to see this once https://deploy-preview-272--openrefine-website.netlify.app/docs/technical-reference/architecture
This is added on the best of my idea, I am open to feedback if any.

PR #272, opened 2023-12-23
For the issue #263 - Add back a top level docs/technical-reference/architecture page to prevent broken links

aakash-a-dev:
Fix for #263

Where on redirecting docs/technical-reference/architecture before this was giving 404 error but now it's gives following output
image
This changes made on the best of my understanding I will appreciate if there is any feedback on which I can work on to make it better.

### DIFF
--- a/docs/technical-reference/Introduction.md
+++ b/docs/technical-reference/Introduction.md
@@ -6,7 +6,7 @@ sidebar_label: Introduction
 
 ### Welcome to the OpenRefine design contribution documentation!
 
-This documentation provides a comprehensive guide to contributing and collaborating on design for OpenRefine. It covers everything from joining the community to creating and implementing design changes.
+This documentation is a comprehensive guide to contributing and collaborating on design for OpenRefine. It covers everything from joining the community to creating and implementing design changes.
 
 The goal of this documentation is to help you make impactful contributions to OpenRefine's user experience. We emphasize the importance of open communication, feedback, and collaboration with developers.
 


magdmartin:
@aakash-a-dev thanks for the PR. The page looks good to me.  Please note that you need to make the reference to the original issue in the PR description (and not the title) for Github to create the link automatically.  I already edited your initial comment to reflect that change. 



@ostephens  is the new page located where you expected to address the 404?

aakash-a-dev:
@magdmartin Thank you, now onwards I will take care of this in mind. I am looking forward to see this as Merge and ready to work on the feedback if there any.

wetneb:
Thanks for working on this! I think we could as well just solve this by adding an actual redirect (not a new page) to either versions.
Redirects can be defined by editing the `static/_redirects` page, which follows the syntax used by Netlify.
https://github.com/OpenRefine/openrefine.org/blob/master/static/_redirects
https://docs.netlify.com/routing/redirects/

If you want to make this a page then I would simply call that "Architecture" (no need to call it a redirect, since it isn't one) and point to both versions in a more concise way.

aakash-a-dev:
@wetneb I did the redirection and tested on Netlify manually it didn't worked so I update the architecture page. If in case we have to do the redirection may be I require a little more direction. But I tried everything mentioned in the documentation.

aakash-a-dev:
Hello @magdmartin and @ostephens I didn't heard on my this PR.

PR #278, opened 2024-01-08
Make date unit table more compact and add missing 'days' unit

wetneb:
The documentation did not mention that `days` (or `day`, `d`) could be used to extract the days part of a date.

Also, I used this opportunity to make the table more compact. Is it clear like this that `days`, `day` and `d` are alternative spellings for the same unit?


tfmorris (approved):
Thanks!

Issue #274, opened 2024-01-08
Migrate to Docusaurus 3.0


thadguidry:
Older versions of Docusaurus will not receive security updates after Jan 31, 2024.
Also, there are many new improvements, here's a few:

* MDX v3
* CommonMark
* Mermaid Mindmap and Quadrant charting
* Debug build
* etc.

and fixes, as well as some breaking changes, but I don't expect us to have hardly any breaking changes.

Press Release:  https://docusaurus.io/blog/releases/3.0

aakash-a-dev:
I went through the documentation, and did a check just 5 more pages remaining to rectified then will be able to see the updated packages in action.

Errors remaining are
[ERROR] 5/114 MDX files couldn't compile!
---
Error while compiling file docs/manual/grelfunctions.md (Line=319 Column=15)
Details: Could not parse expression with acorn
---
Error while compiling file docs/manual/wikibase/configuration.md
Details: These MDX global variables do not seem to be available in scope: batch_id lang version
---
Error while compiling file docs/technical-reference/migrating-older-extensions.md
Details: These MDX global variables do not seem to be available in scope: basedir
---
Error while compiling file docs/technical-reference/openrefine-api.md (Line=198 Column=49)
Details: Could not parse expression with acorn
---
Error while compiling file static/img/2022survey/2022-survey-openanswers.md (Line=544 Column=3)
Details: Expected a closing tag for `<cheeky>` (544:423-544:436) before the end of `paragraph`

PR #279, opened 2024-01-09
Migrate to Docusaurus 3.0 Completed With Checks

aakash-a-dev:
Here, I completed the checkmark and also updated it to Docusaurus 3.0 although there is also 3.1.0 but still I updated to 3.0 for now. 
Before checklist till making it Docusaurus 3.0:
7 errors

After checklist till making it Docusaurus 3.0:
All resolved

Hoping for a positive feedback and positive review for this PR.

Fix: #274


aakash-a-dev:
The error occurring due to node node v16.20.2 (npm v8.19.4), but we have to update the node engine to  "node": ">=18.0", do we require to do this in code, or anyone have to setup this in Netlify side?

aakash-a-dev:
Here is it not accepting my PR, it due to Node engine 18, so I deployed it manually
https://659d918bd0b91930b5787c8a--spectacular-kringle-679929.netlify.app/
Waiting for your follow up on this.

wetneb (changes requested):
Hi @aakash-a-dev, thank you so much for having a go at this! Looking at the "Files changed" tab of this pull request, I see quite some Docusaurus example files which you seem to have committed into your pull request. Needless to say, I think those should not be included. The rest of the changes look promising, I would review them once you have removed those example files.

aakash-a-dev:
@wetneb Thank you for your comment, I went through it. Truly sorry for this. I rectified the example data files uploaded.

aakash-a-dev:
I also corrected the exporting.md when I went through the file changes few lines were unnecessary. Thank you for let me give my attention towards here.

docs/manual/exporting.md
```diff
@@ -4,13 +4,13 @@ title: Exporting your work
 sidebar_label: Exporting
 ---
 
-## Overview {#overview}
+## Overview \{#overview}
```
wetneb, line 7: Any idea why escaping would be needed here but not in other files where the heading ids are also made explicit in the source? For instance, you did not escape them in `docs/manual/grelfunctions.md` but that page seems to still compile fine, no?

docs/manual/wikibase/configuration.md
```diff
@@ -35,7 +35,7 @@ aware of anyone planning to work on this, though.
 
 The manifest is a JSON object describing all the configuration details necessary for OpenRefine to integrate with your Wikibase instance. As an example, here is the manifest of Wikimedia Commons:
 
-```json
+````json
```
wetneb, line 38: What is the difference between starting a code block with four backquotes compared to just three?

wetneb (commented):
Thanks a lot!

wetneb:
I have updated Netlify to use Node.js 20. Next time you make changes to this PR (for instance, solving the merge conflicts with yarn.lock), the site should compile properly.

docs/manual/exporting.md
aakash-a-dev: Yes, went through the documentation of  Docusaurus 3.0 where there is a command to check what will lead error if we upgrade where it shows the problems to the file and I rectified it according to that only. Where in exporting.md I had to do this because while compiling it throws error of react child class can't be rendered, it means it was taking it as a react class so added escape here and it worked.

docs/manual/wikibase/configuration.md
aakash-a-dev: Here, I am sorry. It come's to my practice of doing this because some I use to do it in md files, I will revert it to ``` column only then

docs/manual/exporting.md
aakash-a-dev: I rectified it this time. Thank you and sorry for this feedback.

docs/manual/wikibase/configuration.md
aakash-a-dev: I rectified it this time. Thank you and sorry for this feedback.

aakash-a-dev:
[cause]: Error: Failed to compile due to Webpack errors.
  Error: Docusaurus server-side rendering could not render static page with path /docs/manual/exporting because of error: Objects are not valid as a React child (found: object with keys {}). If you meant to render a collection of children, use an array instead.
  
  Getting this error while building it in netlify. Any way I can rectify it.

docs/manual/exporting.md
ostephens: >so added escape here and it worked.
The `{#overview}` markdown is meant to set a specific anchor name (see https://docusaurus.io/docs/markdown-features/toc#heading-ids). So when you say "it worked" do you mean that there was no longer an error, or did you check that the anchor still works?

ostephens: In any situation where we are explicitly setting an anchor on a heading that would be the same as the automatically generated one, we could remove the explicit setting (see https://docusaurus.io/docs/markdown-features/toc#heading-ids)

wetneb: The intention of adding those anchors was to ensure their stability (to avoid breaking links) when we reword the heading titles. When those headings are written in the source, you can still change them but at least you are doing so consciously.

We did that when preparing the documentation for translation because it was recommended in Docusaurus' docs, I think. Of course, there has not been much progress on the translation front, but we got some feedback that translating the docs would indeed be very useful to some communities (in fact, more useful than translating the UI of the tool, because that makes it harder to relate to tutorials in other languages, or to ask for support in English).

aakash-a-dev: image
I found this error now... Can you help me understand is there any react child?

thadguidry:
I would upgrade to 3.1 then in this case?  ... that way we can take advantage of the main thing folks are complaining about just like us... "broken anchors".  They added a new checker for that in v3.1
> 🚀 New Feature
docusaurus-mdx-loader, docusaurus-module-type-aliases, docusaurus-theme-classic, docusaurus-types, docusaurus-utils, docusaurus
[#9528](https://github.com/facebook/docusaurus/pull/9528) feat(core): make broken link checker detect broken anchors - add onBrokenAnchors config ([@OzakIOne](https://github.com/OzakIOne))

docs/manual/exporting.md
aakash-a-dev: Never mind I found it and I guess this time it will work.

aakash-a-dev:
Now It passed every case!. It worked this time found out the issue.
In export.md on line no 92 ({{ }}) was getting considerate as react child component so put that into ` ` so now it worked with out any conflict.

aakash-a-dev:
@thadguidry I upgraded this to 3.1 only, when I went through documentation that's the best thing I thought to do. So the current code is for 3.1 only.

thadguidry:
@aakash-a-dev but did you read the comment that l linked to?  It specifically addresses @ostephens concern about the header ids, I think?  But it looks like their official docs don't mention those things?  Maybe we should ask them?

> [slorber](https://github.com/slorber) commented [on Dec 14, 2022]?(https://github.com/facebook/docusaurus/issues/3321#issuecomment-1351459064)
> Thanks for reporting @xiaogaozi
> 
> Note as part fo the MDX 2 migration (https://github.com/facebook/docusaurus/pull/8288), {#headingId} is not a valid syntax anymore, and Titus suggests we use {/*#headingId*/} instead.
> 
> It should also be possible to escape ids: \{#headingId}
> 
> Make sure to adapt once we upgrade to MDX 2

ostephens:
> @aakash-a-dev but did you read the comment that l linked to? It specifically addresses @ostephens concern about the header ids, I think? But it looks like their official docs don't mention those things? Maybe we should ask them?
> 
> > It should also be possible to escape ids: /{#headingId}

@thadguidry This is the approach that @aakash-a-dev has taken so all good I think

aakash-a-dev:
> > @aakash-a-dev but did you read the comment that l linked to? It specifically addresses @ostephens concern about the header ids, I think? But it looks like their official docs don't mention those things? Maybe we should ask them?
> > > It should also be possible to escape ids: /{#headingId}
> 
> @thadguidry This is the approach that @aakash-a-dev has taken so all good I think, although feedback to the Docusaurus documentation would make sense

I made it again /{#headingId} to {#headingId} I did it to rectify an issue but comes to know that the issue was taking place due to something else and I rectified it and mentioned in above comment. I guess we don't have to worry about escape ids now.

aakash-a-dev:
Hello I didn't heard from a long time on this PR, so just thought to put a follow up comment. If anywhere you think I require to make modification I am happy to do that. Thank you.

docs/manual/cellediting.md
```diff
@@ -148,7 +148,7 @@ This can help match cells that have typos, or incorrect spaces (such as matching
 
 For an in-depth understanding of N-gram fingerprinting, check this [document](https://github.com/OpenRefine/OpenRefine/wiki/Clustering-In-Depth#n-gram-fingerprint)
 
-_**Phonetic Clustering {#phonetic-clustering}**_
+_**Phonetic Clustering \{#phonetic-clustering}**_
```
wetneb, line 151: Let's give this the same look as the surrounding "headings":
```suggestion
**<a name="phonetic-clustering">Phonetic Clustering</a>**
```

wetneb:
Could you fix the merge conflict with `yarn.lock` please? After that it should be good to merge.

aakash-a-dev:
Pardon! I actually made the changes you asked and also worked on conflict error of yarn.lock file and every thing get done, is there any further changes I require to make? Will happy to that.

wetneb (approved):
Thanks a lot!

PR #290, opened 2024-02-01
Blog post to announce new advisory committee members

wetneb:
@magdmartin I've slightly reformatted and added links to your text, let me know if it works for you

Also I have removed the `slug` in the header: those are not necessary for new blog posts, they were just there for older ones (to preserve URLs).


wetneb:
@magdmartin I would have reduced the resolution of those images as they are rather big. Sadly once the PR is merged, they are in the git repository "forever" unless we force-push to clean things up. It's something worth keeping in mind whenever you add images here.

tfmorris:
I think it's still worth fixing. The second image takes several seconds to download in multiple chunks on my 100+ Mb/s connection. That's not a very good user experience. 

p.s. Even force pushing them won't reduce the repository size. The repo history would need to be rewritten.

magdmartin:
I apologize, I haven't considered the file size.  If you can offer some advice on how to rewrite the repository, I've done some research and it appears to be quite complex with a potential risk of corrupting the repo. I don't feel confident enough to take action myself.

tfmorris:
To clarify, I think it's worth replacing the image file with one of a smaller size and matching cropping, but I don't think it's worth rewriting the git history. The file is 4.5MB, not 100s of MB, and the repo is 23 MB on disk (vs 547 MB for the source code repo).

wetneb:
I reduced the image sizes.

PR #295, opened 2024-02-08
Added Operator Section to GREL Page

aayushk9:
Fixes #289 
Introduced an 'Operator' section to the GREL page, improving user accessibility to information about operators in GREL.

PR #296, opened 2024-02-10
Added Operator Section to GREL Page 

### LINKED ISSUE #289: Documentation of GREL operators is lacking
Although the functionality of operators in GREL is quite intuitive, there is some overloading and special functionality that is not documented in an easy way to find. For example, the overloading of the + operator for string concatenation is only mentioned in the 'Basics' section, while the != (not equal to) operator is not mentioned at all. The special behaviour of the / operator for ints is nicely mentioned at the top of the Math functions section, but the % (integer modulus) operator gets no mention (though it does work - I use it a lot!). It would be great if there were either a separate section on operators, or else that all operators are described in each object type section i.e. at the top of each of 'String functions', and 'Math Functions' (there are no overloaded operators for array types, sadly, so no need for a mention in that section). I'm not aware of any operators that work for Date objects (presumably == and != work, but it might be logical to assume the relational operators like < and > work for these and someone might look to the documentation to find out.

aayushk9:
Fixes #289 
Introduced an 'Operator' section to the GREL page, improving user accessibility to information about operators in GREL.

### DIFF
--- a/docs/manual/grel.md
+++ b/docs/manual/grel.md
@@ -22,6 +22,31 @@ Evaluating conditions uses symbols such as `<`, `>`, `*`, `/`, etc. To check whe
 
 See the [GREL functions page for a thorough reference](grelfunctions) on each function and its inputs and outputs. Read on below for more about the general nature of GREL expressions.
 
+## Operators {#operators}
+
+#### Arithmetic Operators {#arithmetic-operators}
+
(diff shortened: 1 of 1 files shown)


aayushk9:
Hey @wetneb 
Just checking in on the pull request. If you've had a chance to review it, please let me know if there are any changes or adjustments needed.

wetneb (approved):
Per @ostephens' and @tfmorris' review in #295

Issue #275, opened 2024-01-08
Allow PDF file download for offline documentation


thadguidry:
There is a Docusaurus community plugin to produce PDF files for documentation:
https://github.com/signcl/docusaurus-prince-pdf

which is referenced here: https://docusaurus.io/community/resources#community-plugins

This might allow us to generate a PDF version of our docs for users to have (thinking of regular users, training, etc.) as an example to drop to USB drives easily for offline reference.
We could then put a submenu perhaps directly under Documentation nav called "Offline PDF Docs download" or some such.
I think we had 1 or 2 folks ask for this before? dunno.

monstajoe2002:
Hi! Can you assign me this issue?

monstajoe2002:
I just finished with creating the pdf. It was a lot of work, but I got it working here. I used github actions to automate the pdf creation process using `docusaurus-prince-pdf` and upload it as an artifact. There is a catch, however. In order to generate the PDF, we must spin up the local server in the cloud and gh actions are asynchrounous i.e. if a step fails the whole workflow stops unless we state otherwise. The `serve` command required runs fine but it's blocking the execution of the other steps. Therefore, I had to cancel the workflow to execute the remaining steps. Sorry for the long rant. Can I still make pull request here? Or do you have a better solution. Maybe I'm new to Github Actions so idk 😅

wetneb:
That sounds like a pretty complicated solution indeed, I am sorry you spent so much time on it!
Re-generating the PDF file at each build of the website looks too complicated and is probably overkill. We could probably do it once in a while, outside of the CI. It's not like the docs change so often.

So I think it would be perfectly fine to just make a PR adding the Docusaurus plugin without it being added to the continuous deployment pipeline (which is running in Netlify, not GitHub Actions by the way).
We can think afterwards about how to advertise it on the website.

monstajoe2002:
Thanks bro! One thing I should mention about this plugin is that you run it with `npx docusaurus-prince-pdf` without integrating it in the project.

monstajoe2002:
I created a PR for this issue: https://github.com/OpenRefine/openrefine.org/pull/304

magdmartin:
Thanks for the PR @monstajoe2002. Sorry we didn't check the license thoroughly before starting the integration. Upon review of the plugin and Prince License, I realized that the [non-commercial license](https://www.princexml.com/license/) requires us to :
[...]

PR #304, opened 2024-02-28
Upload Offline pdf docs

monstajoe2002:


wetneb:
Apologies in the delay for reviewing this. I don't think uploading a static PDF in this repository is a good move as it will get outdated as docs are edited.

ostephens:
Perhaps implementing a docusaurus plugin like https://github.com/jean-humann/docs-to-pdf might be an option?

monstajoe2002:
I tried that, it's no different than the other plugin. Also, it didn't work on my machine.

ostephens:
@wetneb having looked more closely now (sorry @monstajoe2002 ) I realise that the intent here is that the pdf is generated and updated by a GitHub workflow https://github.com/OpenRefine/openrefine.org/pull/304/commits/2d9bf1aa3705817fd0a3ccc43d83a837bcd168f1

If this works I don't think there should be concerns about it being outdated as long as it gets generated as the repo is updated?

ostephens:
Sorry both - I've deleted my last comment as I think it was totally wrong

ostephens:
@monstajoe2002 I agree with @wetneb uploading a static file is not a good solution here. I feel there must be a better solution - what were the issues you saw with docs-to-pdf and prince?

thadguidry:
A much better way to handle this would be to NOT USE PDF, but instead bundle the website and maintain all the HTML content.

Folks would locally view the site just by unzipping the bundled site and then clicking on the `index` file and the routing is handled through hash routing and the site locally accessed with `file://index.html`.
We could bundle the site (or perhaps just the `/docs` folder)  at build time into a GitHub artifact release and on the site have a link in the footer or header that downloads that offline site zip file.
The implementation would use the new **Docusaurus offline browsing feature** now supported in version 3.4 (which I also have a PR for folks to review that fixes the broken anchors).

More detail on how to do this is here: https://docusaurus.io/blog/releases/3.4#hash-router---experimental

ostephens:
@monstajoe2002 I'd really like to support achieving your goal here, but I wonder if you could:

* describe in more detail what your aim is (why are you adding the PDF - is it for a printable version? Or to have one that works offline?)
* describe the issues you encountered when trying to use the plugins to automatically generate the PDF

magdmartin:
@ostephens I agree with your point. Let's keep the requirement in #275 
@monstajoe2002 I am closing this PR as we refine the requirement first.

PR #311, opened 2024-03-20
Update funding.md

magdmartin:
adding a link to Wikimedia Meta-Wiki for ORefine: 23 Wikimedia


magdmartin:
I checked locally, and the `npm run build` is successful. I can't find how to retry the netlify check without pushing a new commit. Clicking `rerun` under the check tab seems to do nothing.

PR #312, opened 2024-03-20
Update 2023-08-25-reproducibility-designer.md

magdmartin:
I am marking the position as filled so we don't receive new applications.

### DIFF
--- a/blog/2023-12-07-announcing-zoe-cooper.md
+++ b/blog/2023-12-07-announcing-zoe-cooper.md
@@ -1,7 +1,7 @@
 ---
 author: Martin Magdinier
 title: Announcing Zoe Cooper as OpenRefine Lead Designer
-slug: 2023/12/07/announcing-zoe-cooper
+slug: 2023/12/07/announcing-zoe-cooper.html
 ---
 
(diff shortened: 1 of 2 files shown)


wetneb:
Sorry about this failure, Docusaurus isn't so user-friendly. You could try with the full URL instead of a relative one (for which there should be a working syntax, but maybe you have better things to do than to figure out which one).

magdmartin:
The hiring announcement page was missing the slug parameter. I was able to successfully test it locally, but the build still failed.

magdmartin:
@wetneb should be good to merge.

blog/2023-12-07-announcing-zoe-cooper.md
```diff
@@ -1,7 +1,7 @@
 ---
 author: Martin Magdinier
 title: Announcing Zoe Cooper as OpenRefine Lead Designer
-slug: 2023/12/07/announcing-zoe-cooper.md
+slug: 2023/12/07/announcing-zoe-cooper.html
```
wetneb, line 4: Changing the slug changes the URL of the page, which creates dead links. In the future, can we stick to not using any `slug` at all, letting Docusaurus generate the URL by itself? It would avoid this sort of problem…

PR #320, opened 2024-04-10
Update minimum Node version to 20

wetneb:
For https://github.com/OpenRefine/OpenRefine/pull/6526

### DIFF
--- a/docs/technical-reference/build-test-run.md
+++ b/docs/technical-reference/build-test-run.md
@@ -11,7 +11,7 @@ This page explains how to install the tools you need to run OpenRefine from sour
 * [OpenRefine's source code](#get-openrefine-source-code);
 * a [Java Development Kit (JDK)](#set-up-jdk) (version 11 or later);
 * [Apache Maven](#maven);
-* [Node.js and NPM](#nodejs) (version 16 or later).
+* [Node.js and NPM](#nodejs) (version 20 or later).
 
 ### Get OpenRefine's source code {#get-openrefine-source-code}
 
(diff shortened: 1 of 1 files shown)


docs/technical-reference/build-test-run.md
```diff
@@ -147,7 +147,7 @@ Other distributions are likely to offer Maven in their official package reposito
 ### Node.js and npm {#nodejs}
 
 
-The OpenRefine webapp requires [Node.js](https://nodejs.org/en/download/) and npm to install package dependencies. We require Node.js 16 or newer.
+The OpenRefine webapp requires [Node.js](https://nodejs.org/en/download/) and npm to install package dependencies. We require Node.js 20 or newer.
```
tfmorris, line 150:
```suggestion
The OpenRefine webapp requires [Node.js](https://nodejs.org/en/download/) and npm to install package dependencies.
```

Consider dropping version reference or replacing it with something generic like "We generally require the latest LTS version of Node.js."

tfmorris (changes requested):
There's also https://github.com/OpenRefine/openrefine.org/blob/03f4c050b4680bd7e8b60953017e02966426c499/docs/technical-reference/functional-tests.md?plain=1#L51

Even if we can't have everything in a single repo, it would be good to minimize the number of duplicate references. They just get out of date/sync.

docs/technical-reference/build-test-run.md
wetneb: I'd rather keep it here and drop it from the Cypress instructions because someone trying to get Cypress to run has probably already met the dependencies to compile and run OpenRefine anyway.

magdmartin:
Since https://github.com/OpenRefine/OpenRefine/pull/6526 was merged three weeks ago, it will be good to keep the documentation up to date. 
@tfmorris  are we good to go with @wetneb proposal?

docs/technical-reference/build-test-run.md
tfmorris: OK

tfmorris:
Sure, that's fine.

PR #322, opened 2024-04-14
Update grelfunctions.md with Tip for Date functions

thadguidry:
- Adds TIP to mention that Date functions only work with Strings or Dates, not Numbers.  As witnessed with user https://forum.openrefine.org/t/only-year-please/1453?u=thadguidry
image

### DIFF
--- a/docs/manual/grelfunctions.md
+++ b/docs/manual/grelfunctions.md
@@ -447,6 +447,12 @@ As of OpenRefine 3.4.1, uniques() reorders the array items it returns; in 3.4 be
 
 ## Date functions {#date-functions}
 
+:::tip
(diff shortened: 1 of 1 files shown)


tfmorris (changes requested):
This represents a bug in the implementation, so is not something we should document.

There is code in `ToDate` to check for `Long`s and convert them to `String`s, but I suspect perhaps the user data had `Integer`s instead. Whatever the cause, we should fix the bug and add test coverage, rather than documenting it.

thadguidry:
Closing based on @tfmorris assessment.  We can always come back to address updating the docs once bug is fixed.

PR #324, opened 2024-04-16
Add section about update notifications

wetneb:
For https://github.com/OpenRefine/OpenRefine/pull/6511

### DIFF
--- a/src/pages/privacy.md
+++ b/src/pages/privacy.md
@@ -57,6 +57,6 @@ used when importing or exporting projects to a Google Drive, or for Wikibase aut
 
 ## Update notifications on OpenRefine's start page
 
-To detect whether a new version is available, OpenRefine makes web queries to external websites when its start page is opened. This can be disabled by [setting the system property](manual/running#jvm-preferences) `refine.display.new.version.notice` to `false`.
+To detect whether a new version is available, OpenRefine makes web queries to external websites when its start page is opened. This can be disabled by [setting the system property](docs/manual/running#jvm-preferences) `refine.display.new.version.notice` to `false`.
 * Before OpenRefine 3.8, the queries are made to GitHub.com. See [GitHub's privacy policy](https://docs.github.com/en/site-policy/privacy-policies/github-general-privacy-statement) to find out what information gets stored out of those requests.
 * From OpenRefine 3.8 on, the queries are made to openrefine.org, hosted by Netlify. See [Netlify's data processing agreement](https://www.netlify.com/pdf/netlify-dpa.pdf) for details about data they might store. The OpenRefine project does not have access to analytics for those requests, like the rest of the traffic on openrefine.org.
--- a/src/pages/privacy.md
+++ b/src/pages/privacy.md
@@ -55,4 +55,8 @@ encryption in your operating system.
 OpenRefine does not store user data in cookies, except for OAuth authentication,
 used when importing or exporting projects to a Google Drive, or for Wikibase authentication.
 
+## Update notifications on OpenRefine's start page
 
+To detect whether a new version is available, OpenRefine makes web queries to external websites when its start page is opened. This can be disabled by [setting the system property](manual/running#jvm-preferences) `refine.display.new.version.notice` to `false`.
+* Before OpenRefine 3.8, the queries are made to GitHub.com. See [GitHub's privacy policy](https://docs.github.com/en/site-policy/privacy-policies/github-general-privacy-statement) to find out what information gets stored out of those requests.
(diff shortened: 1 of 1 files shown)


wetneb:
I'd merge this soon (alongside with https://github.com/OpenRefine/OpenRefine/pull/6511) unless anyone has concerns.
Obviously the wording can be adjusted afterwards independently of the 3.8 release.

src/pages/privacy.md
```diff
@@ -55,4 +55,8 @@ encryption in your operating system.
 OpenRefine does not store user data in cookies, except for OAuth authentication,
 used when importing or exporting projects to a Google Drive, or for Wikibase authentication.
 
+## Update notifications on OpenRefine's start page
 
+To detect whether a new version is available, OpenRefine makes web queries to external websites when its start page is opened. This can be disabled by [setting the system property](docs/manual/running#jvm-preferences) `refine.display.new.version.notice` to `false`.
+* Before OpenRefine 3.8, the queries are made to GitHub.com. See [GitHub's privacy policy](https://docs.github.com/en/site-policy/privacy-policies/github-general-privacy-statement) to find out what information gets stored out of those requests.
+* From OpenRefine 3.8 on, the queries are made to openrefine.org, hosted by Netlify. See [Netlify's data processing agreement](https://www.netlify.com/pdf/netlify-dpa.pdf) for details about data they might store. The OpenRefine project does not have access to analytics for those requests, like the rest of the traffic on openrefine.org.
```
tfmorris, line 62: We don't have access to our own analytics? That doesn't seem productive. But if this statement is accurate, it's fine for now.

src/pages/privacy.md
```diff
@@ -55,4 +55,8 @@ encryption in your operating system.
 OpenRefine does not store user data in cookies, except for OAuth authentication,
 used when importing or exporting projects to a Google Drive, or for Wikibase authentication.
 
+## Update notifications on OpenRefine's start page
 
+To detect whether a new version is available, OpenRefine makes web queries to external websites when its start page is opened. This can be disabled by [setting the system property](docs/manual/running#jvm-preferences) `refine.display.new.version.notice` to `false`.
```
tfmorris, line 60:
```suggestion
To detect whether a new version is available, OpenRefine makes web queries to external websites when its start page is opened. This can be disabled by [setting the system property](docs/manual/running#jvm-preferences) `refine.display.new.version.notice` to `false`, which is its default value unless the user has opted-in.
```

Hopefull the above is true about the default

tfmorris (changes requested):
Looks fine with the one suggestion about the default privacy setting.

src/pages/privacy.md
thadguidry: We should have access to Algolia search reports however.  Different, but still, useful stuff there to see what people search the site for.  I like Plausible thus far on a few other projects I'm with and is super easy to add in their script and make it work for Docusaurus sites.

src/pages/privacy.md
wetneb: No, that's the contrary.

tfmorris: I forgot about the historical case. I've updated my suggested text. If that disagrees with the code, let's adjust the code so that it actually is opt-in (as it should be).

wetneb: That's still not quite right! There are two things:
* the system property, which if set to `true`, disables the entire feature (including prompting the user to opt in)
* the OpenRefine preference value, which set when the user opts in

That might sound like a complicated set up but I think that's actually fitting: in environments where the user has no control over applying updates, it makes sense for the system administrator to set `refine.display.new.version.notice` to `false` so that the user doesn't get prompted at all. Then, after prompting, it makes sense to store the user's preference in the preference store indeed, so that they are able to change this choice easily.

wetneb: I'm merging this as is so that it's in place for the release, but happy to tweak this part further of course.

src/pages/privacy.md
tfmorris:
```suggestion
* From OpenRefine 3.8 on, the queries are made, only if the user opts in, to openrefine.org, hosted by Netlify. See [Netlify's data processing agreement](https://www.netlify.com/pdf/netlify-dpa.pdf) for details about data they might store. The OpenRefine project does not have access to analytics for those requests, like the rest of the traffic on openrefine.org.
```

src/pages/privacy.md
tfmorris: OK, so on further checking, `refine.display.new.version.notice` (which actually means "don't poll for version updates") appears to have been [added in 3.6](https://openrefine.org/whats_new#new-features-1), so I've updated my suggested text to reflect that.

I've also added a suggestion to the post-3.8 privacy notice to say that it's opt-in.

wetneb: I can't apply your suggestions after merging the PR, so I have copied them over to another commit: https://github.com/OpenRefine/openrefine.org/commit/99a4fff6b2bb53ca02c492122d46ddf70cc90a57. In such a situation it'd be more convenient if you could make the commit directly (or a PR if you prefer)

PR #332, opened 2024-04-22
Fix #331 Meet accessible contrast ratio

### LINKED ISSUE #331: Some blue colored content is too light.  Ensure primary color uses at least WCAG AA 4.5:1 contrast ratio
Our current Nav links and other areas of the documentation when viewed with `data-theme="light"` , the default for `root` is extremely too light and even fails AA 4.5:1 contrast ratio minimum.  (Even for me, I've noticed it with my own eyes being a bit too light).

We should add new appropriate color ranges that at least give our primary color and a few others close to the 4.5:1 contrast ratio for both `light` and `dark` theme.

We want to be at least here after changes are made, where in my example I've manually adjusted the nav bar link color.
But we'll need to do the changes in the `/src/css/custom.css` file fully :
image

thadguidry:
Fix #331

- Adjusts color ranges for Infirma to meet WCAG AA 4.5:1 contrast ratio
- Adds slightly lighter ranges for `dark` theme to meet as well.

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -9,13 +9,13 @@
 
  /* Primary color meets WCAG AA contrast ratio of 4.5:1 */
  :root {
-    --ifm-color-primary: #2272a0;
-    --ifm-color-primary-dark: #1f6790;
-    --ifm-color-primary-darker: #1d6188;
-    --ifm-color-primary-darkest: #185070;
-    --ifm-color-primary-light: #257db0;
-    --ifm-color-primary-lighter: #2783b8;
-    --ifm-color-primary-lightest: #2c94d0;
+    --ifm-color-primary: #005e94;
+    --ifm-color-primary-dark: #005585;
+    --ifm-color-primary-darker: #00507e;
+    --ifm-color-primary-darkest: #004268;
+    --ifm-color-primary-light: #0067a3;
+    --ifm-color-primary-lighter: #006caa;
+    --ifm-color-primary-lightest: #007ac0;
     --ifm-color-secondary: #999999;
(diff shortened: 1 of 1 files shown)


wetneb:
Thanks for this!
It's hard to get an overview of all the color changes but for the one of the download button, the new color strikes me as significantly away from the color palette used in the logo.
For that color, how about using `#0078ff`, which is the darkest shade of blue in the logo? And leaving the other ones untouched? Or if you want to adjust them, perhaps they could only be made darker/lighter, so that the shades of blue remain consistent with the ones from the logo. For instance, there are a couple of colors starting with `#00…` indicating that they have no red component, so I would expect that when making them darker, that component remains `#00…`.

thadguidry:
@wetneb Slightly better now?  I've pulled the range over to more closely match what we had before.  I.E. the color matrix is just slightly darker blues and only `--ifm-color-primary-darkest` fails just slightly the contrast test (but it's close enough).

thadguidry:
There's also a few other changes that I think are generally necessary after this...
For instance, our `<a>` links are really hard to spot as links within paragraphs!  That's where I think we can customize the CSS to have them as
`font-weight: bold;` 

Also when hovering over certain types of elements like `<button>`s etc.  I don't think it's contrasting enough now after this color change?  There also, we can just tweak the CSS for those kinds of elements that just need a slight tweak for `hover { background-color: xxx }` style changes.

Another new issue for those kinds of small tweaks.

thadguidry:
### Darker Blues for better contrast than original lighter blues
image
image
image
image

wetneb (approved):

PR #338, opened 2024-05-24
add refine-js to extensions.md

magdmartin:
Following this post https://forum.openrefine.org/t/live-coding-adding-groovy-as-an-expression-language/1538/3?u=martin I discovered the refine-js extension. 
I am adding refine-js extension


tfmorris (approved):
Looks fine, but we should probably come up with some type of organization scheme for the list, whether it be alphabetical or categorical or ...

That can be done separately though.

thadguidry:
> Looks fine, but we should probably come up with some type of organization scheme for the list, whether it be alphabetical or categorical or ...

Agree.  Maybe have a category column added, and within it we start with
1. Expression languages
2. everything else
3. . . .

wetneb:
Note that it's really just a prototype: vanilla Javascript isn't very useful as an expression language because most useful things in Javascript are provided by external packages, and this doesn't provide a way to install them.
I made it just because it's a fun fact that OpenRefine already embeds a Javascript interpreter, so the extension doesn't need to add any… but as mentioned before on the long term we should get rid of this interpreter, to replace `controller.js` by something declarative.

PR #339, opened 2024-05-26
fix grammar in installing extensions section 

thadguidry:
- we should be more explict and mention the top-level folder for the extensions path.


magdmartin:
Looks good to me

PR #360, opened 2024-08-05
Add admonition about Date functions now using UTC

thadguidry:
- Fixes https://github.com/OpenRefine/OpenRefine/issues/6009

### DIFF
--- a/docs/manual/exploring.md
+++ b/docs/manual/exploring.md
@@ -47,6 +47,10 @@ A “date” type is created when a column is [transformed into dates](transform
 
 Date-formatted data in OpenRefine relies on a number of conversion tools and standards. For something to be considered a date in OpenRefine, it will be converted into the ISO-8601-compliant extended format with time in UTC: YYYY-MM-DDTHH:MM:SSZ.
 
+:::info
+Date/times without timezone info were interpreted as **local** up until May 2018 when OpenRefine 3.0 was released, at which point they were switched from **local** to **UTC**.  One benefit of this change was to introduce consistency and reproducibility when working collaboratively and sharing projects across timezones.  Discussed in issue [#6009](https://github.com/OpenRefine/OpenRefine/issues/6009)
+:::
+
 When you run <span class="menuItems">Edit cells</span> → <span class="menuItems">Common transforms</span> → <span class="menuItems">To date</span>, the following column of strings on the left will transform into the values on the right:
 
 |Input|→|Output|
--- a/docs/manual/exploring.md
+++ b/docs/manual/exploring.md
@@ -48,7 +48,7 @@ A “date” type is created when a column is [transformed into dates](transform
 Date-formatted data in OpenRefine relies on a number of conversion tools and standards. For something to be considered a date in OpenRefine, it will be converted into the ISO-8601-compliant extended format with time in UTC: YYYY-MM-DDTHH:MM:SSZ.
 
 :::info
-Date/times without timezone info were interpreted as **local** up until May 2018 when OpenRefine 3.0 was released, at which point they were switched from **local** to **UTC**.  One benefit of this change was to introduce consistency and reproducibility when working collaboratively and sharing projects across timezones.  Discussed in issue [#6009](https://github.com/OpenRefine/OpenRefine/issues/6009)
+All datetimes are stored internally as UTC with full date & time precisions. During conversion from strings, date/times without timezone info were interpreted as **local** up until May 2018 when OpenRefine 3.0 was released, at which point they were switched from **local** to **UTC**.  One benefit of this change was to introduce consistency and reproducibility when working collaboratively and sharing projects across timezones.  Discussed in issue [#6009](https://github.com/OpenRefine/OpenRefine/issues/6009)
 :::
 
 When you run <span class="menuItems">Edit cells</span> → <span class="menuItems">Common transforms</span> → <span class="menuItems">To date</span>, the following column of strings on the left will transform into the values on the right:
--- a/docs/manual/exploring.md
+++ b/docs/manual/exploring.md
@@ -47,10 +47,6 @@ A “date” type is created when a column is [transformed into dates](transform
 
 Date-formatted data in OpenRefine relies on a number of conversion tools and standards. For something to be considered a date in OpenRefine, it will be converted into the ISO-8601-compliant extended format with time in UTC: YYYY-MM-DDTHH:MM:SSZ.
 
-:::info
-All datetimes are stored internally as UTC with full date & time precisions. During conversion from strings, date/times without timezone info were interpreted as **local** up until May 2018 when OpenRefine 3.0 was released, at which point they were switched from **local** to **UTC**.  One benefit of this change was to introduce consistency and reproducibility when working collaboratively and sharing projects across timezones.  Discussed in issue [#6009](https://github.com/OpenRefine/OpenRefine/issues/6009)
-:::
-
 When you run <span class="menuItems">Edit cells</span> → <span class="menuItems">Common transforms</span> → <span class="menuItems">To date</span>, the following column of strings on the left will transform into the values on the right:
 
 |Input|→|Output|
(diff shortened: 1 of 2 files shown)


thadguidry:
@tfmorris and @ostephens Do you think this would close that linked issue?

docs/manual/exploring.md
```diff
@@ -47,6 +47,10 @@ A “date” type is created when a column is [transformed into dates](transform
 
 Date-formatted data in OpenRefine relies on a number of conversion tools and standards. For something to be considered a date in OpenRefine, it will be converted into the ISO-8601-compliant extended format with time in UTC: YYYY-MM-DDTHH:MM:SSZ.
 
+:::info
+Date/times without timezone info were interpreted as **local** up until May 2018 when OpenRefine 3.0 was released, at which point they were switched from **local** to **UTC**.  One benefit of this change was to introduce consistency and reproducibility when working collaboratively and sharing projects across timezones.  Discussed in issue [#6009](https://github.com/OpenRefine/OpenRefine/issues/6009)
```
tfmorris, line 51:
```suggestion
All datetimes are stored internally as UTC with full date & time precisions. During conversion from strings, date/times without timezone info were interpreted as **local** up until May 2018 when OpenRefine 3.0 was released, at which point they were switched from **local** to **UTC**.  One benefit of this change was to introduce consistency and reproducibility when working collaboratively and sharing projects across timezones.  Discussed in issue [#6009](https://github.com/OpenRefine/OpenRefine/issues/6009)
```

docs/manual/grelfunctions.md
```diff
@@ -447,6 +447,10 @@ As of OpenRefine 3.4.1, uniques() reorders the array items it returns; in 3.4 be
 
 ## Date functions {#date-functions}
 
+:::info
+Date/times without timezone info were interpreted as **local** up until May 2018 when OpenRefine 3.0 was released, at which point they were switched from **local** to **UTC**.  One benefit of this change was to introduce consistency and reproducibility when working collaboratively and sharing projects across timezones.  Discussed in issue [#6009](https://github.com/OpenRefine/OpenRefine/issues/6009)
```
tfmorris, line 451: The most relevant place for this information is with the `toDate()` and `toString()` documentation, although those should probably be brief references to a more general discussion.

docs/manual/grelfunctions.md
```diff
@@ -447,6 +447,10 @@ As of OpenRefine 3.4.1, uniques() reorders the array items it returns; in 3.4 be
...
+:::
+
 ###### now() {#now}
 
 Returns the current time according to your system clock, in the [ISO 8601 extended format](exploring#data-types) (converted to UTC). For example, 10:53am (and 00 seconds) on November 26th 2020 in EST returns [date 2020-11-26T15:53:00Z].
```
tfmorris, line 456: Not related to this PR, but since it caught my eye, ISO 8601 isn't involved. Although this does the same thing as before, due to the changes in `toString()`, the effective date/time is different.

tfmorris (changes requested):
Thanks for tackling this! I was hoping someone was going to address documenting the incompatible changes.

I think the description needs fleshing out a little bit with a general discussion of the change and compatibility implications, as well as specific references near the affected functions (mostly `toDate()` and `toString()`, I think).

The current proposal only mentions benefits without any drawbacks. Is that intentional? It seems to me to leave out half the story.

docs/manual/grelfunctions.md
thadguidry: I don't quite understand.  This is the `now()` function directly, which returns a date-time with an offset from UTC/Greenwich in the ISO-8601 calendar system, according to the Java docs.  So I think this reads OK as is?  But I think your point is that we need to improve the `toString()` documentation more fully?  I can do that in this PR also.

docs/manual/grelfunctions.md
thadguidry: I wanted to have this generally at the beginning of discussing all the Date functions.
But would love your help on the `toString()` documentation suggestions which I can add in this PR.

docs/manual/grelfunctions.md
thadguidry: Actually, that's wrong... `now()` will always return the date-time from the users system clock in the specified time-zone (which is `Z` [according to our code](https://github.com/OpenRefine/OpenRefine/blob/ebae7adef85322ddd53b749e10fde8eda8bcd8f7/main/src/com/google/refine/expr/functions/date/Now.java#L48) , which means Zulu time, which means we return the users system time without any offset.)

So should `now()` docs stay as is, because it does return the users system clock in ISO 8601 format?

tfmorris: It's only semantics, but the date handling is complicated enough without leading people astray. ISO 8601 dates are represented using strings. What the `now()` function returns is an internal OpenRefine datetime object. The square brackets around "[date 2020-11-26T15:53:00Z]" and the leading "date" prefix indicate that this is the default `toString()` rendering for an internal date object, not a string. The expression `now().toString() would give "2020-11-26T15:53:00Z" for the above example. Using OpenRefine 2.8, `now()` returns a date object rendered as "[date 2024-08-20T17:16:42Z]" while `now().toString()` returns "Aug 20, 2024" using my default system locale and now().toString("hh:mm a z") return "05:20 PM EDT" which matches the current local wall clock time.

In addition to date to string conversions, import/export from spreadsheet formats like Excel and OpenOffice behave differently as well.

thadguidry: Wait, so you want to document that OpenRefine prior to 3.0 handled this differently?  If that is your concern, I think it's less important how prior 3.0 handled things and just document how they currently work over the last few versions.  Our surveys show that everyone is using at least OpenRefine 3.7+ ?

tfmorris: The whole point of #6009 is documenting (instead of fixing) the date handling regression. If that's not what this is addressing, you should remove the tag. 

The discussion of `now()` is separate. It doesn't return an ISO 8601 formatted string. It returns a datetime.

thadguidry: @tfmorris Can you express how you would word things?
>I think the description needs fleshing out a little bit with a general discussion of the change and compatibility implications, as well as specific references near the affected functions (mostly toDate() and toString(), I think).

docs/manual/grelfunctions.md
thadguidry: Moved admonition about date/times regression to section toString()

thadguidry:
OK, I think this is now ready for final review and merging.
I've fulfilled all of Tom's suggestions:
- moving admonition about date/time handling changes since 3.0 into only toString() section
- clean up `now()` removing ISO 8601 mention and giving new examples as he suggested.

thadguidry:
@wetneb We can close the OpenRefine issue #6009 if you can review and agree with my last comment and we merge this? Looks good?

docs/manual/grelfunctions.md
```diff
@@ -56,7 +56,13 @@ You can use toString() to convert numbers to strings with rounding, using an [op
 
 You can also convert dates to strings, using date parsing syntax built into OpenRefine (see [the toDate() function for details](#todateo-b-monthfirst-s-format1-s-format2-)). For example, `value.toString("MMM-dd-yyyy")` would convert the date value [2024-10-15T00:00:00Z] to “Oct-15-2024”.
 
-Note: In OpenRefine, using toString() on a null cell outputs the string “null”.
+:::note
+In OpenRefine, using `toString()` on a null valued cell such as `value.toString()` outputs the string “null”.
+:::
```
wetneb, line 61: I think two boxes in the same go is a bit much emphasis. We got the feedback that we use too much of those boxes in the documentation (which I agree with), so I would rather leave the existing note as is.

thadguidry:
@wetneb Better? I reverted that 1 note.

wetneb (approved):
Thanks! To me, it's still a bit weird to have this in the `toString()` function, as it's not the most intuitive place to learn about date handling in OpenRefine (`toDate()` would be more natural to me) but it's up to you.

thadguidry:
Let's leave it under `toString()` , Tom had a good point that the changes he did actually fall under `toString()` changes anyways, so that's where any regession would be impacted the most, which is where I think his logic was going.

So let's merge please and move on to other important things.

PR #363, opened 2024-08-16
Document levenshteinDistance function in grelfunctions.md

zyadtaha:


docs/manual/grelfunctions.md
```diff
@@ -591,6 +591,9 @@ Returns a boolean indicating whether o has a member field called [name](expressi
 ###### coalesce(o1, o2, o3, ...) {#coalesceo1-o2-o3-}
 Returns the first non-null from a series of objects. For example, `coalesce(value, "")` would return an empty string “” if `value` was null, but otherwise return `value`.
 
+###### levenshteinDistance(s1, s2) {#levenshteinDistance-s1-s2}
```
ostephens, line 594: I think this should go in the `String functions` section of this documentation. I would suggest directly after the `length` function

docs/manual/grelfunctions.md
```diff
@@ -591,6 +591,9 @@ Returns a boolean indicating whether o has a member field called [name](expressi
...
+Returns how many single-character changes you need to make to turn one string into another. This includes adding characters, removing characters, and changing characters to match between the two strings. For example, `levenshteinDistance("New York", "newyork")` will return 1 and `levenshteinDistance("M. Makeba", "Miriam Makeba")` will return 5
```
ostephens, line 595: We should be clear what is returned. So e.g. "Returns an integer indicating how many single-character ..."

ostephens (changes requested):

docs/manual/grelfunctions.md
zyadtaha: Done.

ostephens (approved):

docs/manual/grelfunctions.md
```diff
@@ -41,6 +41,9 @@ Uses the logical operator XOR (exclusive-or) on two or more booleans to output a
 
 Returns the length of string s as a number.
 
+###### levenshteinDistance(s1, s2) {#levenshteinDistance-s1-s2}
+Returns an integer indicating how many single-character changes you need to make to turn one string into another. This includes adding characters, removing characters, and changing characters to match between the two strings. For example, `levenshteinDistance("New York", "newyork")` will return 1 and `levenshteinDistance("M. Makeba", "Miriam Makeba")` will return 5
```
tfmorris, line 45: The examples make it look like the comparison is case insensitive. If so, this should be mentioned, along with any other preprocessing which is done (e.g. diacritic folding, removal of punctation and controls, etc).

zyadtaha: Hi @thadguidry,
Do you have any details on preprocessing methods for the Levenshtein distance in the Simile Vicino implementation?

thadguidry: The only preprocessing that the Levenshtein class calls is through `StringWrapper prepare()` in order to begin computing the `score()`, where it basically caches info about the tokenized version of a string (a [BagOfTokens](https://github.com/TeamCohen/secondstring/blob/master/src/com/wcohen/ss/BagOfTokens.java)).  It basically consists of tokenizing the string and assigning weights.  There's no other preprocessing that I can tell looking through Cohen's code.
You can use an outline view in your IDE to get a sense of how things are connected up from this file: https://github.com/TeamCohen/secondstring/blob/master/src/com/wcohen/ss/api/StringWrapper.java
where SecondString (com.wcohen.ss.*) is the library that Simile Vicino uses for Levenshtein.

tfmorris: The preprocessing currently consists of:
- trim leading and trailing spaces (where "space" is anything less than or equal to Unicode code point  U+0020 using Java's String.trim()
- remove all punctuation and controls - regex \p{Cntrl}|\p{Punct} -> "" (for ASCII definitions of this character classes, not Unicode aware)
- collapse all whitespace to a single space - p{Space}+ -> " " (also ASCII only)
- lowercase using Java's locale sensitive String.toLower() with the locale that the OpenRefine server was started with (typically the default locale of the user logged in)

It does NOT do any diacritic folding.

It should probably at least be made Unicode aware for whitespace and punctuation and perhaps (optionally) do diacritic folding.

tfmorris: Here's the normalization/pre-processing currently:
https://github.com/OpenRefine/simile-vicino/blob/0ba545cb4fe3d835647110eb615e8d1075e78d13/src/edu/mit/simile/vicino/NGramTokenizer.java#L36-L42

It is *not* Unicode aware for its handling of whitespace or punctuation, which probably should be fixed. We may also want to add (optional) diacritic folding. The case folding is done in a locale-sensitive manner using the locale that the OpenRefine server was started with.

zyadtaha: I've updated the documentation accordingly—thanks a lot! I also think adding an option for diacritic folding would be a great enhancement.

PR #364, opened 2024-08-19
Add tip admonition for Wikidata Help:Dates

thadguidry:
- as a result of this issue:  https://github.com/OpenRefine/OpenRefine/issues/6798

### DIFF
--- a/docs/manual/wikibase/schema-alignment.md
+++ b/docs/manual/wikibase/schema-alignment.md
@@ -198,10 +198,6 @@ if the language or the text are invalid).
 
 ### Dates {#dates}
 
-:::tip
-See also https://www.wikidata.org/wiki/Help:Dates
-:::
-
 Dates are parsed from cell contents (or from any constant provided in
 the schema) and the precision of the date is inferred from its format.
 Here are the valid formats:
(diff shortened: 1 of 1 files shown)


tfmorris (commented):
Since there are only two legal values for calendar, perhaps we can just list them near line 224-224 instead of redirecting the user. The Wikidata pages says:
> Currently two calendar models are supported: [proleptic Gregorian calendar (Q1985727)](https://www.wikidata.org/wiki/Q1985727) and [proleptic Julian calendar (Q1985786)](https://www.wikidata.org/wiki/Q1985786)

thadguidry:
Hi @tfmorris , I thought about that. But there are Phabricator issues for Wikidata where eventually Date datatypes will get more love and also the documentation improved.  There's also corner cases and special handling that is covered in their docs.  So I'd rather just direct users to read up on the current Date datatypes on Wikidata's official help pages, which are actually quite hard to find just navigating around and often missed.

wetneb:
It feels weird to me to have a "See also" block at the beginning of the subsection. If it's "see *also*", it makes more sense to me to have this at the end of this subsection, so that it complements the main content.

wetneb (approved):

PR #365, opened 2024-08-19
Add tip admonition for Wikidata Help:Data_type#globe-coordinate

thadguidry:


tfmorris (commented):
In addition to being restricted to Earth, my reading of the docs is that coordinates are also restricted to the WGS84 coordinate system, so perhaps that should be mentioned as well.

thadguidry:
@tfmorris As with the Dates issue, I'd rather users read official Wikidata help pages for themselves.  It's actually better if folks learn more about Wikidata whenever they are updating values in Wikidata.

PR #368, opened 2024-08-30
Add user-friendly guide for custom clustering functions in cellediting.md

zyadtaha:


wetneb (commented):
Thank you,  that looks very good to me!

When referring to button names or other such elements, you can use `<span class="menuItems">Add new function</span>` to make it clearer.

Also, it would be worth making sure that the capitalization you use in this documentation is the same as in the UI (for instance "Add new function", not "Add New Function").

In terms of content, I think we could find more relatable examples for the sample expressions given in the documentation. How about the text below? Maybe it goes already a bit too much in the direction of a tutorial, but given the complexity of this feature perhaps that's okay?

> Say you want to cluster song titles in English which sometimes differ because of included or omitted articles (such as "the" or "a"). For instance, you could have values such as `The Peanut Vendor`, `Peanut Vendor (The)` and `Peanut Vendor`. OpenRefine's default fingerprinting function successfully maps `The Peanut Vendor` and `Peanut Vendor (The)` to the same value `peanut the vendor` (by removing punctuation and re-ordering words), but `Peanut Vendor` is mapped to `peanut vendor`, failing to cluster it with the other values. You could use a custom keying function which filters out articles from the fingerprint value: `filter(value.fingerprint().split(" "), word, (word != "the").and(word != "a").and(word != "an")).join(" ")`. This will map all three values to the same clustering key `peanut vendor`.
> 
> Nearest neighbors with the Levenshtein distance counts the number of changes between two values, but does not relate this number of changes to the lengths of strings. For instance `Fong` and `Fang` have distance 2, but `Sinéad O'Connor` and `Sinéad M. O'Connor` have distance 3. When clustering sets of strings with widely varying lengths, it can be useful to normalize this number of changes by the lengths of the strings. This can be achieved with a custom distance such as `levenshteinDistance(value1, value2) / (max(length(value1), length(value2)) + 1)`. With such an expression, the distance between `Fong` and `Fang` becomes 0.33, whereas `Sinéad O'Connor` and `Sinéad M. O'Connor` are at distance 0.16.

zyadtaha:
> In terms of content, I think we could find more relatable examples for the sample expressions given in the documentation. How about the text below? Maybe it goes already a bit too much in the direction of a tutorial, but given the complexity of this feature perhaps that's okay?

Sounds good! I agree that more relatable examples can make the feature easier to understand. Thanks for the suggestion! 

> For instance `Fong` and `Fang` have distance 2,

I think you mean distance 1?

Additionally, `Fong` and `Fang` are not included in the clustering results with the expression you provided, even when using the "native" edit distance clustering method. This occurs because `Fong` and `Fang` do not trigger the `compute` function which computes the distance. Specifically, each of them fails to be tokenized by `this._tokenizer.tokenize(s);` in the `populate` function of the `NGramClusterer.class`, resulting in an empty `tokens` array:
```
    public void populate(String s) {
        Token[] tokens = this._tokenizer.tokenize(s);
        Token[] var3 = tokens;
        int var4 = tokens.length;

        for(int var5 = 0; var5 < var4; ++var5) {
            Token t = var3[var5];
            String ss = t.getValue();
            Set<String> l = null;
            if (!this.blocks.containsKey(ss)) {
                l = new TreeSet();
                this.blocks.put(ss, l);
            } else {
                l = (Set)this.blocks.get(ss);
            }

            ((Set)l).add(s);
        }

    }
```

wetneb:
> I think you mean distance 1?

I computed the distance with the GREL function and it returned 2, so I assumed it did not support replacing characters, only inserting and deleting. Feel free to adjust as needed!

> Additionally, Fong and Fang are not included in the clustering results with the expression you provided, even when using the "native" edit distance clustering method.

Interesting! I have to admit I didn't run clustering for this example but only computed the distances manually with GREL. What you describe sounds like a bug, perhaps? If it is by design (which could also be understandable, meaning that small strings don't get clustered at all), then we could tweak the examples so that they still have clearly different lengths (something like `John Fong` / `John Fang` / `Doctor Sinéad O'Connor` / `Sinéad M. O'Connor` maybe?)

zyadtaha:
> I computed the distance with the GREL function and it returned 2, so I assumed it did not support replacing characters, only inserting and deleting. Feel free to adjust as needed!

Maybe you have some spaces? I tried `levenshteinDistance("Fang", "Fong")` now again and it returns 1.

zyadtaha:
> What you describe sounds like a bug, perhaps?

mmm, maybe we can ask @tfmorris? the native levenshtein clustering do not return `fang` and `fong` in the same cluster even the difference between them is 1?

tfmorris:
I've created https://github.com/OpenRefine/OpenRefine/issues/6839 to describe the problem with the nearest neighbor clusterer.

docs/manual/cellediting.md
```diff
@@ -194,6 +194,33 @@ PPM (Prediction by Partial Matching) uses compression to see whether two values
...
+
+Nearest neighbors with the Levenshtein distance counts the number of changes between two values, but does not relate this number of changes to the lengths of strings. For instance `John Fong` and `John Fang` have distance 1, but `Sinéad O'Connor` and `Doctor Sinéad O'Connor` have distance 7, although they are intuitively more similar. When clustering sets of strings with widely varying lengths, it can be useful to normalize this number of changes by the lengths of the strings. This can be achieved with a custom distance such as `levenshteinDistance(value1, value2) / (max(length(value1), length(value2)) + 1)`. With such an expression, the distance between `John Fong` and `John Fang` becomes 0.1, whereas `Sinéad O'Connor` and `Doctor Sinéad O'Connor` are at distance 0.3.
```
wetneb, line 220: The example got less convincing, right? The original example I picked was so that the relative ordering between the scores would flip depending on whether or not the distance is normalized.
Would it not make sense to use `Doctor Sinéad O'Connor` and `Doctor Sinéad M. O'Connor`?

wetneb: and maybe `J. Fong` / `J. Fang` because it's probably not enough…

zyadtaha: > The original example I picked was so that the relative ordering between the scores would flip depending on whether or not the distance is normalized.

Could you please expand on this? not sure I fully get it.

zyadtaha: Oh, I see now! I initially thought you chose these examples to ensure that the score between `Sinéad O'Connor` and `Sinéad M. O'Connor` would be less than 1, so they would be clustered together. But it looks like the main reason was actually to demonstrate how the relative ordering between the scores of the two pairs would flip.

docs/manual/cellediting.md
```diff
@@ -194,6 +194,33 @@ PPM (Prediction by Partial Matching) uses compression to see whether two values
...
+
+Nearest neighbors with the Levenshtein distance counts the number of changes between two values, but does not relate this number of changes to the lengths of strings. For instance `Fong` and `Fang` have distance 1, but `Sinéad O'Connor` and `Sinéad M. O'Connor` have distance 3, although they are intuitively more similar. When clustering sets of strings with widely varying lengths, it can be useful to normalize this number of changes by the lengths of the strings. This can be achieved with a custom distance such as `levenshteinDistance(value1, value2) / (max(length(value1), length(value2)) + 1)`. With such an expression, the distance between `Fong` and `Fang` becomes 0.2, whereas `Sinéad O'Connor` and `Sinéad M. O'Connor` are at distance 0.16.
+
+For more details on built-in clustering functions, see [Clustering In Depth](https://github.com/OpenRefine/OpenRefine/wiki/Clustering-In-Depth).
```
wetneb, line 222:
```suggestion
For more details on built-in clustering functions, see [Clustering In Depth](https://openrefine.org/docs/technical-reference/clustering-in-depth).
```

wetneb (approved):

PR #379, opened 2024-09-19
Clarify Regular expression syntax paragraphs

thadguidry:
- this clarifies a few differences and uses italics for emphasis and introduces the notion that our Regex engine is Java-based, but the syntax is Javascript-based.


magdmartin:
It looks good to me. I will merge this in 48 hours if there is no other feedback.

PR #386, opened 2024-10-19
Security announcement for 3.8.3

wetneb:
Given that we are going to patch severe vulnerabilities in 3.8.3, I think it could be worth pre-announcing the release. I've picked 24th October 2024 as release date, but it can be changed of course.

### DIFF
--- a/blog/2024-10-21-announcing-version-3-8-3.md
+++ b/blog/2024-10-21-announcing-version-3-8-3.md
@@ -1,0 +1,7 @@
+---
+author: Antonin Delpeuch
+title: "Important security fixes in upcoming 3.8.3 version"
+---
+
+On Thursday 24th October 2024, we will publish version 3.8.3 of OpenRefine, with fixes to a collection of important security vulnerabilities.
+Given the severity of those vulnerabilities, we encourage users to get ready to upgrade swiftly to this new version.


thadguidry:
@tfmorris @wetneb I thought that since we are also introducing at least 1 new feature (along with security patches) that we are supposed to update the minor version `.8` (and not only patch version `.3`)?

wetneb:
@thadguidry which new feature do you expect this to introduce? For me this will only contain security patches.

thadguidry:
Any of these https://github.com/OpenRefine/OpenRefine/issues?q=is%3Aclosed+milestone%3A3.8
that are:
1. not a bug
2. not previously part of 3.8.2 ?

But I struggle to see the forest through the trees because we don't do well in our milestone minor/patch labeling or which would automatically add value as a pseudo-autocreated roadmap.

wetneb:
> Any of these https://github.com/OpenRefine/OpenRefine/issues?q=is%3Aclosed+milestone%3A3.8
> that are:
> 1. not a bug
> 2. not previously part of 3.8.2 ?

I'm not sure why anyone would have added an issue to the 3.8 milestone if it's not part of 3.8.2… that would be confusing indeed! Why do you think this would happen? I don't understand what the motivation would be, especially for new features: it wouldn't make sense to add them to the 3.8 milestone, because they should be released as 3.9, right? I am very confused!

> But I struggle to see the forest through the trees because we don't do well in our milestone minor/patch labeling which would automatically add value as a pseudo-autocreated roadmap.

As far as I can see @tfmorris has put some effort to add things to the 3.9 milestone. Does that help?
I think we can totally think about improvements to our workflows though.

thadguidry:
@wetneb Sure, just bringing up the same point as in the barcamp, that having alignment somehow on a roadmap, even for development, is useful for all.  My stance is that we should manage the milestones better.  I see other projects using them better, and accouting for minor/patch and sometimes even nano to denote ABI backward compatibility ( like 3.8.3.1) which we don't need to worry about that level, but still we can have many more milestones to generally just organize better and at least for development - have a clear picture of what the dev roadmap looks like, even if it might change and we move issues and features into further later milestones.  We all agreed in the barcamp to at least get better at milestones, as a start, then we can later decide how to publish a roadmap better for users - based on our better tracking with milestones.

wetneb:
Sure - it's probably a discussion for somewhere else then

tfmorris:
If the confusion was caused by the items at the top of this sorted list:
https://github.com/OpenRefine/OpenRefine/issues?q=is%3Aclosed+milestone%3A3.8+sort%3Aupdated-desc

There are definitely some things which were added to the 3.8 **milestone** recently, but nothing was added to the 3.8 **branch** or **releases**. As part of the preparation for the 3.9-beta1 release ([draft release notes](https://github.com/OpenRefine/OpenRefine/releases/tag/untagged-27099d87a558594d452b)), I made a pass through all issues closed in the last year or two which weren't tagged with a milestone and which weren't flagged as Duplicates or Won't Fix and made an attempt to assign them to the correct milestone.

I was primarily interested in untagged 3.9 issues, but I also:
- tagged as 3.8 those issues who's fixes I verified were on the 3.8 branch
- reopened issues which said they were fixed in 4.0, but which only existed on the `4.0` branch and not `master` (note that this did NOT include any issues tagged with the 4.0 milestone since I was only looking at issues with no milestone - we still need to deal with all those)

Note that this is primarily a matter of convenience for easy reference. The official record is the commit history on the relevant branch.

> My stance is that we should manage the milestones better. 

Feel free to contribute -- or rally other volunteers to contribute. This administrative work is not something a release manager needs to do (unless it hasn't been done by the time the release is due).

>  I see other projects using them better, for example https://github.com/apache/airflow/milestones as just one

That project had 71 active developers in the past week. It is about the furthest from comparable to OpenRefine that one would be able to find.

> We all agreed in the barcamp to at least get better at milestones,

Probably in one of the sessions scheduled at 4am. Do you have a link?

thadguidry:
@tfmorris https://forum.openrefine.org/t/openrefine-2024-barcamp-strategizing-openrefine-roadmap/1633

OK, my takeaway from this discussion is that you'd rather not do as much detailed project planning, and for other contributors to help with that load.  Perfect, that's all I needed to hear.  Antonin has already been vocal before that he'd prefer to do much less administrative work, which is also perfectly fine.

I don't mind helping with planning and organizing towards goal posts and leveraging milestones and projects for keeping track of work needed towards goal posts - as I mentioned just now in a comment in that forum thread - moving the discussion into that forum thread now.  Bye bye now from this PR ;-)

tfmorris:
@thadguidry to close out the part that's relevant to this thread, can you confirm whether or not there was a misunderstanding about the contents of the 3.8 release stream? I haven't added any new features to it and I don't think @wetneb has either.

thadguidry:
@tfmorris ah yes, sorry.  You cleared up the confusion.

PR #391, opened 2024-10-29
Add a link to OR’s CLI in the Automating section

antoine2711:
Adding in Automating section a link to OR’s CLI from Felix Lohmeier.

### DIFF
--- a/docs/manual/running.md
+++ b/docs/manual/running.md
@@ -541,6 +541,7 @@ The following are all third-party extensions and code; the OpenRefine team does 
 
 Some examples:
 
+* A Command Line Interface (CLI) for OpenRefine that enables scripts to automate OpenRefine, by Felix Lohmeier: [orcli](https://github.com/opencultureconsulting/orcli)
 * This project allows OpenRefine to be run from the command line using [operations saved in a JSON file](#reusing-operations): [OpenRefine batch processing](https://github.com/opencultureconsulting/openrefine-batch)
 * A Python project for applying a JSON file of operations to a data file, outputting the new file, and deleting the temporary project, written by David Huynh and Max Ogden: [Python client library for Google Refine](https://github.com/maxogden/refine-python)
 * And the same in Ruby: [Refine-Ruby](https://github.com/maxogden/refine-ruby)


wetneb (approved):

tfmorris:
Wouldn't it be better to remove mentions of all these unsupported clients, since they could be considered implicit endorsements? They are using an undocumented, unsupported, and, most importantly, **untested** internal protocol which has very limited error handling/reporting (because it's designed to be used in a restricted web client with user supervision).

thadguidry:
@tfmorris I think we do a good job in the prior 2 paragraphs of explaining caveats, and "not responsible" attitudes well enough? No?  But perhaps you'd like to suggest something additional into those paragraphs to make it more clear "not an endorsement", and "undocumented, untested, internal protocol used with limited error handling"?

antoine2711:
> Wouldn't it be better to remove mentions of all these unsupported clients, since they could be considered implicit endorsements?

@tfmorris: I think they should go to a different page.

Personally, I would had another section, like “External tools and services”, and there, I would do a page for the clients (the section I’m modifying now), a subsection or a page for all the extentions, and a subsection/page for the Reconciliations Services.

They would all benefit from more explanation for the purpose, installation, usage, etc.

Regards, Antoine

thadguidry:
@antoine2711 Just throwing this out there, but... for things outside the control of OpenRefine, perhaps another alternative of where that information could be posted or captured into would be Wikibooks or Wikiversity articles?

@magdmartin and I had discussed places where the community might begin doing those sorts of things, like writing up things about OpenRefine and its ecosystem.  Rather than us hosting a Wiki ourselves to allow the community to use, I felt that [Wikibooks](https://meta.wikimedia.org/wiki/Wikibooks) (textbooks or manuals) or [Wikiversity](https://meta.wikimedia.org/wiki/Wikiversity) (manuals or training resources, articles) might be better and save our team from maintenance pain?

wetneb:
In my opinion the existing warning is already pretty scary, I don't see a need to deter people from using those tools even more than what we currently do.

I also think we should re-evaluate this policy of discouraging people from using the HTTP API externally. The HTTP API is legitimately used by our own extensions - as seen with the issue created by removing CORS support from the get-rows command. It's difficult for me to understand why we'd decide to be so conservative for changes to Java classes (everything being part of the stable API, even the most obscure classes) and have no stability to offer on the frontend side.

magdmartin:
Should the documentation page redirect to our current listing on the main website: https://openrefine.org/extensions#client-libraries? This would avoid maintaining a list in two different places. We could also reproduce the warning on that page.  


Those libraries are actively, used. The preliminary results of the 2024 user survey show that about 7% of our users rely on one of those libraries (16 out of 226).

thadguidry:
@magdmartin Agree, I'd suggest we move that small listing and caution under Automation.  "Running" is just not semantically the same as "Using it with...(i.e. automating, etc.)".  That would leave the Automation section on `running.md` to be very lightweight paragraph where we can then provide a link to the `extensions.md` clients-libraries section.

antoine2711:
> This would avoid maintaining a list in two different places.

@magdmartin : why should the documentation about tools, extensions and services be out of our current documentation? What’s the advantage for a user to have to look at 2 different places? That why I’m proposing to had a third section in our documentation. Here’s the link to the forum discussion about that: [[Improving the documentation with a new section, “External tools and Services”](https://forum.openrefine.org/t/improving-the-documentation-with-a-new-section-external-tools-and-services/1864)](https://forum.openrefine.org/t/improving-the-documentation-with-a-new-section-external-tools-and-services/1864/3)

Regards, Antoine

PR #392, opened 2024-10-29
Updates testing guidelines for Cypress

thadguidry:


docs/technical-reference/functional-tests.md
```diff
@@ -120,17 +123,19 @@ See below on the dedicated section 'Testing utilities'
 
 ### Testing guidelines {#testing-guidelines}
 
+- Sometimes things are hidden until hovered, like Facet choices. Use `.invoke('trigger', 'mouseover')` since we use jQuery. See [Cypress docs - invoke Examples](https://docs.cypress.io/api/commands/invoke#Properties-that-are-functions-are-invoked)
+- OpenRefine uses jQuery and sometimes you might need to assert a jQuery element method. See [Cypress docs - jQuery method](https://docs.cypress.io/api/commands/invoke#jQuery-method)
```
tfmorris, line 127: These seem like pretty specific cases. If we're going to keep them, they should probably move to the end of the list, after the more general recommendations like not using wait.

tfmorris (changes requested):
Thanks! Just one suggestion...

PR #393, opened 2024-10-30
Update running.md to Launch OpenRefine from the Terminal on MacOS

antoine2711:
Update running.md to Launch OpenRefine from the Terminal to make work in MacOS 14.7 Sonoma.


antoine2711:
@tfmorris: as a mac user, I would particularly appreciate having your feedback on this change.

Regards, Antoine

wetneb (approved):
Approving by default

PR #396, opened 2024-10-31
Update releases.json for 3.8.5

tfmorris:


wetneb (approved):
Thanks! For what it's worth I've done those sorts of changes without PRs, in the interest of having the website updated swiftly after a release.

PR #397, opened 2024-10-31
Update release process based on recent run

tfmorris:
- put changeable portion of mvn command at the end
- remove obsolete link to wiki
- add a note about draft -> publish flakiness

### DIFF
--- a/docs/technical-reference/version-release-process.md
+++ b/docs/technical-reference/version-release-process.md
@@ -7,16 +7,16 @@ sidebar_label: How to do an OpenRefine version release
 When releasing a new version of Refine, the following steps should be followed:
 
 1. Make sure the `master` branch is stable and nothing has broken since the previous version. We need developers to stabilize the trunk and some volunteers to try out `master` for a few days.
-2. Change the version number in [RefineServlet.java](http://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/RefineServlet.java#L62) and in the POM files using `mvn versions:set -DnewVersion=2.6-beta -DgenerateBackupPoms=false`. Commit the changes.
+2. Change the version number in [RefineServlet.java](http://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/RefineServlet.java#L62) and in the POM files using `mvn versions:set -DgenerateBackupPoms=false -DnewVersion=3.9-beta`. Commit the changes.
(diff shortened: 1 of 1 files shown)


docs/technical-reference/version-release-process.md
```diff
@@ -7,16 +7,16 @@ sidebar_label: How to do an OpenRefine version release
...
 3. Compose the list of changes in the code and on the wiki. If the issues have been updated with the appropriate milestone, the Github issue tracker should be able to provide a good starting point for this.
 4. Tag the release candidate in git and push the tag to Github. For example:
 ```shell
 git tag -a -m "Second beta" 2.6-beta.2
     git push origin --tags
 ```
-5. Create a GitHub release based on that tag, with a summary of the changes as description of the release. Publishing the GitHub release will trigger the generation of the packaged artifacts, which will be added to the release on GitHub.
+5. Create a GitHub release based on that tag, with a summary of the changes as description of the release. Publishing the GitHub release will trigger the generation of the packaged artifacts, which will be added to the release on GitHub. Note that if you create a draft release and then publish it, it may not trigger the release workflow on Github Actions.
```
thadguidry, line 17:
```suggestion
5. Create a GitHub release based on that tag, with a summary of the changes as description of the release. Publishing the GitHub release will trigger the generation of the packaged artifacts, which will be added to the [releases on GitHub](https://github.com/OpenRefine/OpenRefine/releases). Note that if you create a draft release and then publish it, it may not trigger the release workflow under our [Github Actions](https://github.com/OpenRefine/OpenRefine/actions).
```

Also, previously, we show how to tag.  But here, we're handwaving a bit and only saying to "Create a GitHub release based on that tag..."  We should describe the process itself a bit better through either/both:
- 1. a `gh release create v3.9.0 ` cli command line example (ref: https://cli.github.com/manual/gh_release_create)
- 2. link(s) or simply describe steps of the "create release" process like they should go to a specific page like our [GitHub Releases -> Draft a new release](https://github.com/OpenRefine/OpenRefine/releases/new)

thadguidry (changes requested):

docs/technical-reference/version-release-process.md
wetneb: I think the singular is correct and "trigger the release workflow on Github Actions" is more natural than "under our Github Actions"

Issue #376, opened 2024-09-16
Date function 'u' Day Num in week, returning year


pumalife:
value: 2024-01-17T00:00:00Z
value.toString('u')
should return the day number for the current week ( ie 1-7) as indicated in https://openrefine.org/docs/manual/grelfunctions#date-functions

but its returning 2024
ive tried this with a text value and converted it to date, manually specifying the format date text value. ie value.toDate('y/m/d').toString('u') and it still outputs 2024

SymbolofMoon:
Hi @tfmorris ,

I have gone through the docs and found out that 
E returns the day in Text (i.e in Tuesday) and e returns the number of week (2)

You can check the below screenshot.

![screenshot-docs_oracle_com-2024_11_15-01_23_36](https://github.com/user-attachments/assets/ee88de95-2c04-47e5-994d-693cd131fe3f)

Also please let me know if the issue is open for contribution

thadguidry:
I'll have to defer to Tom on this particular issue, so that we get it 100% correct.

tfmorris:
@SymbolofMoon Sorry I missed your question. Thank you for offering to contribute.
@mynk2611 Thank you for contributing to the documentation and fixing the reported problem.

I think there are additional documentation updates needed because we no longer use `SimpleDateFormat` for both parsing and formatting. Instead, we use `SimpleDateFormat` on the parsing side of things (ie `toDate()`) and `DateFormatterFormat` on the string formatting (ie `toString()`) side of things. I'll need to do some investigation to see whether we need to fix the code, the docs, or both.

By way of background, this incompatible change to the code was made back in 2018 before the release of OpenRefine 3.0, but the documentation was never updated.

PR #402, opened 2024-11-22
fix documentation for toString() data fromat codes in grelfunctins.md #376

mynk2611:
this is the PR for issue #376 

Screenshot 2024-11-22 at 12 12 38 PM

The documentation is correct now after comparing the correct field letters here: 
https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html


wetneb (approved):
Thank you!

Issue #373, opened 2024-09-11
Call out warning on the installation page are intimidating 


magdmartin:
This issue follows the Heuristic Analysis done by [asengupta](https://forum.openrefine.org/u/asengupta) on [the forum](https://forum.openrefine.org/t/conducting-a-ux-audit/1694/10?u=martin)

The call-out warning on https://openrefine.org/docs/manual/installing can make the process feel high stake or intimidating.

Possible Solution
* integrate call-out warnings with the text
* Use neutral language to not scare the user
* Use consistent pattern for warning 
   * type of content that's body content vs warning
   * language templates
   * consistent styles (font size, weight, case, icon, color)

aakash-a-dev:
I wasn't active on this repo from long time. I am on it. Although it will take a lot of iterations to understand that how warnings will be on the installation page.

magdmartin:
@aakash-a-dev, welcome back! 
This issue primarily concerns copyediting and enhancing the layout of the documentation page at https://openrefine.org/docs/manual/installing and not coding. From the suggested approaches which one are you considering?

aakash-a-dev:
Thank you for clarifying! I plan to focus on **integrating call-out warnings seamlessly into the text** while maintaining clarity and consistency. My approach will include:

1. **Neutral and User-Friendly Language:**  
   I will rephrase the warnings to make them less intimidating while ensuring the necessary emphasis is retained.

2. **Establishing Consistent Styles:**  
   I'll create a consistent pattern for presenting warnings, using uniform styles for font size, weight, iconography, and color. This will ensure warnings are noticeable but not alarming.

3. **Distinguishing Body Content from Warnings:**  
   I'll identify areas where warnings can be merged naturally into the surrounding text or clearly separated, depending on the context. 

4. **Language Templates:**  
   I'll propose standard templates for warnings, so future documentation follows the same structure and tone.

I’ll start by drafting a few examples on how this could look on the installation page and share them for feedback. Iterative refinements will follow based on community input.

---

**Correct Approach for the Issue:**

1. **Review the Current Page:**  
   Examine the existing installation page to identify where the call-out warnings appear overly stark or intimidating.

2. **Design Language Guidelines:**  
   Create guidelines for:
   - **Warning Integration:** Combine warnings with body text where appropriate.
   - **Language Neutrality:** Use constructive phrases (e.g., "Ensure you..." instead of "Warning! Failure to...").
[...]

PR #403, opened 2024-11-23
This is the first draft in hope of getting green flag to continue this or not

aakash-a-dev:
As per discussion in fix: #373 
I had created a very minor first draft in case to reflect my understanding towards the issue. I will need a little more discussion on this but please review this changes.

Before: 
image

image

After: 
image
image

Thank you, I know this is not enough but for an understanding demonstration I am raising this PR.


wetneb:
I think those changes are a clear improvement and would be worth merging as they are. Happy to wait for more if you prefer.

aakash-a-dev:
It's lovely, I will do as per my best of my knowledge. I will raise few more PR afterwards.

wetneb (approved):

PR #409, opened 2024-11-28
Update releases.json to 3.8.7

### LINKED ISSUE #408: Latest release not available on the download page
https://openrefine.org/download lists 3.8.6 as the latest while the latest version is 3.8.7. 3.8.6 also appears to be broken according to users.

magdmartin:
closes #408

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1,8 +1,8 @@
 [
     {
-        "version": "3.8.6",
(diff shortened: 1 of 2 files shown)


tfmorris (changes requested):
Approved for merge with the requested change.

PR #410, opened 2024-12-03
refresh statistic on usage page

magdmartin:
I refreshed the statistics as of December 3, 2024, in preparation for the year-end blog post. 

I changed the format of the Contribution Statistics graph to represent the average number of contributors per year because the previous graph did not display all the values correctly. For the time series, Mermaid required each x value to be explicitly listed, which made the x-axis too crowded when displaying monthly data points. I also tried using a rolling 6-month average, but the axis was still cluttered. I believe that the yearly average is sufficient to demonstrate the trends in contributions over time.


magdmartin:
I will merge it tomorrow if there is no objection

PR #427, opened 2025-02-02
Update to Docusaurus 3.7

### LINKED ISSUE #426: Trying to update to Docusaurus 3.7.0 with React 19 causes issue with Mermaid plugin when using Yarn but not NPM
I've done the necessary research to resolve the issue we would prevent us from upgrading Docusaurus to 3.7.0 and keep using the Mermaid plugin.
https://github.com/facebook/docusaurus/issues/10645

The basic remedy is just to switch over to `npm` instead of `yarn`.  Simple enough.

I think logically this makes sense for a number of reasons:
- npm itself has gotten better and has equivalent features now with yarn.
- nearly every package maintainer ensures things work with npm, but not always 100% with yarn. :-(
- because of the above, we'll get better error info when making forward changes as packages migrate also to React 19 as needed.  Yarn doesn't always give the correct messaging in errors or deprecation warnings (likely because of package maintainers not accounting for the way Yarn handles things).
- Slober himself doesn't use `yarn` and their CI e2e workflow for Yarn PnP doesn't cover the Mermaid theme and thus Docusaurus will likely stay more congruent with `npm`. https://github.com/facebook/docusaurus/issues/10645#issuecomment-2461912194

# TASKS TO DO FOR THIS ISSUE
- [ ] Remove `yarn.lock` file
- [ ] Update to Docusaurus 3.7.0 and React 19.0.0 in `package.json`
- [ ] Update the `netlify.toml` to use `npm` instead of `yarn` commands
- [ ] Update the README.md with instructions for `npm install` `npm start`, etc. instead of `yarn` commands.

thadguidry:
Fixes #426
- changes from yarn to npm
- updates README.md
- updates Docusaurus config and dependencies


thadguidry:
For now, because of https://github.com/algolia/docsearch/issues/2407
we'll need to `--force` the npm dependencies.

Hmm, now how do we do that with Netlify commands (and modify our `netlify.toml` file)?
Reading... https://docs.netlify.com/configure-builds/troubleshooting-tips/#build-fails-on-peer-dependency-conflict

thadguidry:
@magdmartin @ostephens @antoine2711 Does the Netlify preview look OK?  Pages that use Mermaid work OK?

@wetneb Pages where you might have used custom React still work OK?

README.md
```diff
@@ -3,100 +3,56 @@ OpenRefine.org website
...
+Most changes you make to pages and content are reflected live without having to restart the server.
 
 ```sh
-yarn up
+npm run start
 ```
 
-which will update all dependencies and store them in the `yarn.lock` file to also be committed.
-
-### Next version of OpenRefine docs
```
thadguidry, line 62: @wetneb @ostephens  FYI, since we don't do versioning of docs any longer, then this section needed to be removed as well.

antoine2711:
> @magdmartin @ostephens @antoine2711 Does the Netlify preview look OK? Pages that use Mermaid work OK?

@thadguidry: from what I see, it works fine. But I don't know what Mermaid is.

Regards, Antoine

thadguidry:
@antoine2711 @magdmartin seems like Mermaid is only used on usage page and our history page, both in the footer?

thadguidry:
OK, this is ready to merge.
We'll have to keep the `--force` NPM flag on until Algolia releases the updated `@docsearch/react` bump.  But they have fixed and closed the issue.

README.md
```diff
@@ -3,100 +3,56 @@ OpenRefine.org website
...
-corepack enable
-```
-
-Update Yarn to the latest version:
-
-```sh
-yarn set version stable
-```
+- [Node.js (LTS recommended)](https://nodejs.org/en/download/)
+  - includes npm package manager
```
wetneb, line 12:
```suggestion
Building the website requires [Node.js (LTS recommended)](https://nodejs.org/en/download/), which includes the `npm` package manager.
```

PR #449, opened 2025-04-17
Set version to 3.9.3

SoryRawyer:
Now that version 3.9.3 has build artifacts attached to the GitHub release, I am updating the releases JSON file to point to the new latest version.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1,8 +1,8 @@
 [
     {
-        "version": "3.9.2",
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Looks great! Feel free to commit that directly to master, I would say.

SoryRawyer:
Thanks! I had a feeling that would be alright, but I figured I'd open a PR for the first one.

PR #455, opened 2025-05-14
Update goal-posts regarding extension management

magdmartin:
Merge Better testing utilities for extension development with Better support for OpenRefine extensions
@SoryRawyer @wetneb @Abbe98 let me know if it makes sense to merge those two goal post or if we should keep them separated.


SoryRawyer (approved):
I think this makes sense, especially at this stage where both goal posts need a solution definition. It feels appropriate to have "better testing utilities" as a requirement for better support for extensions rather than a standalone goal post.

Abbe98:
I wouldn't personally merge these as I imagine the work involved would be very different. The first point clearly needs work and consensus building in OpenRefine core. For the one about test-ability however, I would imagine that the vast majority of the work would happen outside of OpenRefine core.

Arguably, openrefine/CommonsExtension#157 and related PRs implemented a lot of the second goal post, so we should probably update the description.

SoryRawyer:
The reason behind my approval is that I think the "better support for extensions" goal post will already involve multiple streams of work, and support for testing extensions falls under the umbrella of providing better overall support for extension developers. That said, I don't have a strong opinion on merging these two and I agree that it's possible for some of the current goal posts to grow so much as to become unwieldy.

tfmorris:
@magdmartin Is this still relevant? If so, can  you please resolve the merge conflicts and ping Abbe98 again for review. If he doesn't reply, we can go ahead and merge since it sounds like you think you addressed his feedback.

magdmartin:
@Abbe98 @tfmorris  since I haven't heard back on this I will merge it next week as I am refreshing the goal post page in preparation of the 2026 user survey.

PR #456, opened 2025-05-17
docs(writing extensions): remove advice on how to include  images

Abbe98:
This advice seems outdated. CSS should more or less only be used for decorative images and it's up to extension developers  to include images, in their HTML, JavaScript, or CSS. OR core uses all of the variants.


Abbe98:
@SoryRawyer I completely agree and once I'm done with my work on the JSDoc annotations I imagine I would port that into this documentation.

PR #461, opened 2025-06-04
extensions list: add OpenRefine Hex Colors

Abbe98:


magdmartin:
Thanks for developing and sharing it !

PR #467, opened 2025-06-25
List phonetic algorithms with their correct names

b2m:
The `phonetic` function accepts a limited list of algorithm names. To avoid confusion the names in the documentation now match the expected names by the function.


b2m:
> but I'm not sure if we want to mention `fingerprint` and `ngram-fingerprint`, since they're also available directly as functions

That's what I was thinking about too but decided against it.

Mentioning them would (currently) be technical correct, but then you also would have to explain the difference between `value.phonetic("ngram-fingerprint")`, `value.ngramFingerprint(n)` and `value.ngram(n)`, without adding (much) value to the reader of the documentation.

tfmorris (approved):
Works for me!

PR #487, opened 2025-09-23
update list of client libraries

felixlohmeier:
- merging the list from tech docs with the list from extensions website
- remove latest compatible version info for current client libraries (too many changes needed)
- sort list alphabetically by language
- php client library was deleted by owner on github and packagist


docs/technical-reference/openrefine-api.md
```diff
@@ -283,29 +283,5 @@ The result array will hold up to ten results, depending on how many rows there a
...
-
-### PHP {#php}
-* [openrefine-php-client](https://github.com/keboola/openrefine-php-client)
-
-### Java {#java}
-* [refine-java](https://github.com/ancore/refine-java)
-
-### Bash {#bash}
-* [bash-refine.sh](https://gist.github.com/felixlohmeier/d76bd27fbc4b8ab6d683822cdf61f81d) (templates for shell scripts)
+Libraries using the OpenRefine API are listed on the [Extensions](https://openrefine.org/extensions#client-libraries) website.
```
SoryRawyer, line 287: Would you mind changing this (and the link back to this page on the extensions page) to remove `https://openrefine.org/`? That way, the links will be generated relative to the existing domain so deployments like the Netlify preview will direct this to the Netlify preview version of the extensions page.

SoryRawyer (changes requested):
Thanks for cleaning this up! I just have a small nitpick but otherwise I think this looks great.

felixlohmeier:
I manually checked all changed links in the Netlify deploy preview. Everything should be fine now.

SoryRawyer (approved):
Thanks for the extra work to trim the internal links!

magdmartin:
Thanks for updating the list @felixlohmeier It looks good to me.

src/pages/extensions.md
```diff
@@ -59,24 +59,24 @@ OpenRefine can connect to several reconciliation services. You can find informat
 
 ## Client libraries
 
-You can automate some OpenRefine operations using one of the existing libraries. Those libraries are using the [OpenRefine API](https://github.com/OpenRefine/OpenRefine/wiki/OpenRefine-API).
+You can automate some OpenRefine operations using one of the existing libraries. Those libraries are using the [OpenRefine API](/docs/technical-reference/openrefine-api).
```
tfmorris: "You can automate" is an invitation to users to shoot themselves in the foot. We need a strong warning here about the risks involved.

```suggestion
There are a number of unsupported client libraries which use the internal [OpenRefine client/server protocol](/docs/technical-reference/openrefine-api). Because these use an unsupported protocol which has no tests, these libraries should only be used by technically sophisticated developers who understand the risks and limitations. These libraries are not provided or supported by the OpenRefine team.
```

(I know that this is existing wording and not something changed by this PR)

PR #502, opened 2026-01-02
Updates Docusaurus to 3.9.2

thadguidry:
- and prepares for Docusaurus 4.0.0
- and removes yarn (too many issues with Berry and PnP, and known issues with Docusaurus with yarn if not using node_modules https://github.com/facebook/docusaurus/issues/10645 )

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -48,7 +48,6 @@ return {
           position: 'left',
         },
         {
-          to: '/2025-fundraising',
           to: 'https://store.openrefine.org/',
           label: 'Store',
           position: 'right',
--- a/docs/manual/grelfunctions.md
+++ b/docs/manual/grelfunctions.md
@@ -251,7 +251,9 @@ Takes two strings and compares them, returning a string. Returns the remainder o
 
 ###### escape(s, s mode) {#escapes-s-mode}
 
-Escapes s in the given escaping mode. The mode can be one of: "html", "xml", "csv", "url", "javascript". Note that quotes are required around your mode. See the [recipes](https://github.com/OpenRefine/OpenRefine/wiki/Recipes#question-marks--showing-in-your-data) for examples of escaping and unescaping.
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -8,33 +8,27 @@ Opening a pull request for your changes will generate a preview on Netlify that
...
-Building the website requires [Node.js (LTS recommended)](https://nodejs.org/en/download/), which includes the `npm` package manager.
+- [Node.js (LTS recommended)](https://nodejs.org/en/download/)
 
 ### Installation
 
-Once you have installed Node, which includes the npm package manager, we can install and set-up the dependencies:
+CD into the repository parent folder and install dependencies
 
 ```sh
+cd /openrefine.github.com
```
tfmorris:
```suggestion
cd openrefine.org
```

docusaurus.config.js
```diff
@@ -224,7 +230,7 @@ return {
         blog: {
           blogSidebarCount: 'ALL',
           onInlineAuthors: 'ignore',
-          
+          onUntruncatedBlogPosts: 'ignore'
```
tfmorris: These are only warnings and seem like a useful nudge to the blog post writers to remember to include the truncation markers. What's the rationale for disabling the warnings?

package.json
```diff
@@ -13,26 +13,25 @@
...
+    "react-dom": "19.2.3",
     "react-json-view-lite": "^2.3.0",
     "react-lite-youtube-embed": "^2.3.1",
     "ua-parser-js": "^1.0.37"
   },
   "engines": {
     "node": ">=18.0"
-  }
+  },
+  "packageManager": "yarn@4.12.0"
```
tfmorris:
```suggestion
  }
```

The PR comment says "removes Yarn" but this looks like it is actually *adding* Yarn. Let's keep that separate from the version updates. It's also looks like it could be the reason that CI/deploy is failing.

docusaurus.config.js
thadguidry: I think @magdmartin didn't like the trunaction markers.  So, ???  Not sure what to do here, up to both of you.

package.json
thadguidry: ah, let me fix that!

PR #504, opened 2026-01-06
fix: improve button text contrast in dark mode (#503)

### LINKED ISSUE #503: Dark mode: Button text color has insufficient contrast
When dark mode is enabled, button backgrounds remain blue (as expected), but the button text color changes to black.

This results in poor contrast and reduced readability, as the text should remain light-colored (e.g. white) against the blue background, similar to light mode.

Expected:
Button text color maintains sufficient contrast in dark mode (e.g. white text on blue background).

Actual:
Button text appears black in dark mode while the background remains blue.

This issue may be related to the broader theming work (#7585), but is scoped to a specific button text color inconsistency.

Screenshots attached.

Image

Image

I’d be happy to work on a small fix for this, if that sounds reasonable.

Harshit-Mourya:
### Summary
Fixes an accessibility issue where primary button text appeared black in dark mode,
resulting in insufficient contrast against the blue background.

### Changes
- Ensure button text uses a light color in dark mode for better readability
- Align dark-mode button text contrast with light-mode behavior

#### Before (Dark mode)
Screenshot 2026-01-04 121623

#### After (Dark mode)
Screenshot 2026-01-06 103824

### Result
- Improved visual clarity in dark mode
- Better contrast and readability
- Improved accessibility without changing button background colors

Fixes #503

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -188,5 +188,5 @@ code {
 /* Fix button text contrast in dark mode */
 html[data-theme="dark"] .button.button--primary,
 html[data-theme="dark"] .button.button--secondary {
-  color: var(--ifm-color-white);
+  color: var(--ifm-font-color-base);
 }


src/css/custom.css
```diff
@@ -3,183 +3,190 @@
...
+@media only screen and (max-width: 600px) {
+  #magic {
+    height: 100px;
+  }
+}
+
+/* Fix button text contrast in dark mode */
+html[data-theme="dark"] .button.button--primary,
+html[data-theme="dark"] .button.button--secondary {
+  color: var(--ifm-color-white);
```
tfmorris: This white is fairly stark in contrast to the rest of the page. Perhaps we can use something like 
```suggestion
  color: var(--ifm-font-color-base);
```
if it provides enough contrast.

tfmorris (changes requested):
Thanks for the contribution. I suggest we use a slightly less bright white, if it provides enough contrast.

Also, for future contributions, it makes it much easier to review if the changes are limited to just those focused on the fix without mixing in white space and formatting changes.

Harshit-Mourya:
Thanks for the suggestion, @tfmorris !

Agreed — pure white is a bit stark in dark mode. I’ll switch the button text to `var(--ifm-font-color-base)` and confirm contrast is still sufficient.

Also noted the feedback about keeping PRs tightly scoped. I’ll push the update shortly.

PR #511, opened 2026-01-17
fix broken anchors and cleanup markdown syntax in places

thadguidry:


docs/manual/columnediting.md
```diff
@@ -26,60 +26,62 @@ New columns will be named after the original column, with a number: “Location
...
 
-The simplest way to use this operation is simply leave the default `value` in the expression field, to create an exact copy of your column. For a column of [reconciled data](reconciling), you can use the variable `cell` instead, to copy both the original string and the existing reconciliation data. This will include matched values, candidates, and new items. 
+The simplest way to use this operation is simply leave the default `value` in the expression field, to create an exact copy of your column. For a column of [reconciled data](reconciling), you can use the variable `cell` instead, to copy both the original string and the existing reconciliation data. This will include matched values, candidates, and new items.
 
 One useful expression is to create a column based on concatenating (merging) two other columns. Select either of the source columns, choose <span class="menuItems">Edit column</span> → <span class="menuItems">Add column based on this column...</span>, name your new column, and use the following format in the expression window:
 
-```
+```grel
```
tfmorris: Does our Docusaurus flavored Markdown understand this language label? I'm guessing it doesn't out of the box, but perhaps our config has been tweaked somehow?

docs/manual/wikibase/schema-alignment.md
```diff
@@ -127,33 +127,33 @@ reference will be discarded but the reference will still be added
...
-  mode will delete all statements with a particular main property on the target entity.
-* **Property and value**, which compares statements by their main property and main value only. This is what QuickStatements does. In addition, it is possible (and enabled by default) to match statement values in a lax way, for instance to ignore
-  differences in trailing whitespace or rounding of quantities.
-* **Qualifiers**, which compare statements using their property, main value and qualifiers. It is possible to define a list of property identifiers which determines which qualifiers are discriminating. Other qualifiers will not be taken into account when
-  comparing statements. By default, all qualifiers are taken into account. This matching strategy also supports lax value matching.
+
+- **Property**, which compares statements by their main property only. This means that any two statements using the same main property will be considered equivalent. For intance, using this merging strategy in conjunction with the **Delete** editing mode will delete all statements with a particular main property on the target entity.
```
tfmorris:
```suggestion
- **Property**, which compares statements by their main property only. This means that any two statements using the same main property will be considered equivalent. For instance, using this merging strategy in conjunction with the **Delete** editing mode will delete all statements with a particular main property on the target entity.
```

tfmorris (changes requested):
Thanks for all the updates! That's a lot and I admit I was kind of skimming by the end, but looks good. 

I've got one question and a few suggestions, the latter of which come from Code Pilot, but for some reason it didn't generate suggestions for them.

I'm not sure why Netlify hasn't deployed the updates, but I'm happy to merge this when it's been updated and sort out the issues, if any, later.

docs/manual/columnediting.md
tfmorris: Thanks for the explanation. That makes sense.

PR #524, opened 2026-02-04
Move translation and documentation contributor pages up a level in contributor sidebar

### LINKED ISSUE #523: Add sidebar items for other contribution types
The [contributor "getting started" documentation](https://openrefine.org/docs/technical-reference/contributing) lists four contribution types. However, only "developer" and "designer" have top-level items in the "Contributing to OpenRefine" section of the sidebar. Given that we have documentation for the other two types, I think it would make sense to move those pages to the sidebar along "developer" and "designer".

Specifically, I'm imagining:
- The [page to translate OpenRefine's interface](https://openrefine.org/docs/technical-reference/translating-ui) would move up one level on the sidebar and be named "Contributing as a translator"
- The [page on documentation contributions](https://openrefine.org/docs/technical-reference/documentation-contributions) would be added to the sidebar as an item named "Contributing as a documentor" (I'm also fine with changing the pattern of these section names, but I think it would be nice if they're consistent)

SoryRawyer:
fixes #523

Moves translation and documentation pages into the sidebar under "Contributing to OpenRefine" so that they align with the code and design contribution types. This PR also renames the sidebar labels from translation and documentation contributions to match the naming for the other contribution types. Finally, the documentation page is updated to remove stale documentation and point to the repo README to keep documentation in one place.

### DIFF
--- a/docs/technical-reference/documentation-contributions.md
+++ b/docs/technical-reference/documentation-contributions.md
@@ -1,7 +1,7 @@
 ---
 id: documentation-contributions
 title: Documentation contributions
-sidebar_label: Contributing as a documenter
+sidebar_label: Contributing as a technical writer
 ---
 
 We welcome contributions to our documentation, which is maintained as a set of [Markdown](https://www.markdownguide.org/) files in the [OpenRefine/openrefine.org](https://github.com/OpenRefine/openrefine.org) Git repository.


docs/technical-reference/documentation-contributions.md
```diff
@@ -1,7 +1,7 @@
 ---
 id: documentation-contributions
 title: Documentation contributions
-sidebar_label: Documentation contributions
+sidebar_label: Contributing documentation
```
tfmorris: I agree with the comment in the original issue that it would be preferable to have all four sidebar entries have consistent wording. Was it intentional that this wasn't done?

docs/technical-reference/translating-ui.md
```diff
@@ -1,7 +1,7 @@
 ---
 id: translating-ui
 title: Translate OpenRefine's interface
-sidebar_label: Translate OpenRefine's interface
+sidebar_label: Contributing translations
```
tfmorris: Same comment as above about consistent wording.

SoryRawyer:
Thanks for the review! Here is a screenshot with the sidebar text updated:
Screenshot from 2026-02-04 21-18-20

magdmartin (commented):
Overall, I think this is a step in the right direction and can help prompt future contributors to improve the documentation regarding those two roles.

PR #550, opened 2026-03-05
Remove outdated Ant and org.json references from writing-extensions.md

### LINKED ISSUE #522: Remove outdated code references from extension documentation
The [documentation on writing extensions](https://github.com/OpenRefine/openrefine.org/blob/master/docs/technical-reference/writing-extensions.md) references a few outdated parts of the codebase. One example is the [menu bar section](https://github.com/OpenRefine/openrefine.org/blob/master/docs/technical-reference/writing-extensions.md#main-menu-main-menu), which references a JS file that no longer exists.

nishitagajraj:
Removes outdated references to the Ant build system (build.xml) and old org.json library (JSONObject, JSONWriter) which were replaced in OpenRefine 3.1 and 3.2 respectively.

Fixes #522

### DIFF
--- a/docs/technical-reference/writing-extensions.md
+++ b/docs/technical-reference/writing-extensions.md
@@ -93,7 +93,7 @@ sample-extension/
 
 The sub-directory `MOD-INF` contains the Butterfly module's metadata and is what Butterfly looks for when it scans directories for modules. `MOD-INF` serves similar functions as `WEB-INF` in other web frameworks.
 
-Java code is built into the sub-directory `classes` inside `MOD-INF`, and supporting external Java jars are in the `lib` sub-directory. Those will be automatically loaded by Butterfly. (The build.xml script is wired to compile into the `classes` sub-directory.)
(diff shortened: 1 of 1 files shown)


docs/technical-reference/writing-extensions.md
```diff
@@ -315,15 +315,15 @@ Packages.com.google.refine.model.Project.registerOverlayModel(
...
+static public OverlayModel reconstruct(ObjectNode o) throws IOException {
       ...
   }
 ```
 
 When the project gets saved, the overlay model instance's `write` method will be called:
 
 ```
-public void write(JSONWriter writer, Properties options) throws JSONException {
+public void writeJSON(ObjectMapper writer) throws JsonProcessingException {
```
tfmorris: Thanks for identifying the obsolete reference, but the update isn't a one-to-one replacement. You'll need to check the new code to figure out the proper documentation.

tfmorris (changes requested):
Thanks for reviewing and updating the documentation! I think it needs a little more work, but it's a good start.

PR #576, opened 2026-06-06
  fix(reconciliation-api): update spec links to W3C final report URLs

sponomarev72:
The GitHub-hosted spec pages at reconciliation-api.github.io are no longer
the canonical location. This replaces those links with the official W3C
Community Group final report URLs. Also fixes one link that incorrectly
pointed to v0.1 instead of v0.2.


magdmartin:
Thanks - it looks good to me.

PR #593, opened 2026-07-27
update homepage and event notification for 2026 user survey

magdmartin:
As part of the https://forum.openrefine.org/t/drafting-the-2026-user-survey/2822/

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -25,6 +25,9 @@ function HeaderSection() {
           <img src="img/openrefine_logo.svg" alt="OpenRefine logo" style={{maxWidth: "min(250px, 20vw)", display: 'block', margin: 'auto'}}/>
         </div>
       </div>
+     <div>
+      <Link className="button button--primary" to="https://forms.gle/EAqF1uYyDNsoaPyz8" style={{fontSize: '1.5em'}}>💎 Participate to the 2026 user survey ✨</Link>
+    </div>
       <div className="row">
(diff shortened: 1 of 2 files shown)


Abbe98 (approved):
The button breaks the layout on medium and small screens but it looks like it has been the case for the last few years. 

Better getting this out there sooner than later but lets keep the issue in mind for the future.

src/pages/index.tsx
```diff
@@ -25,6 +25,9 @@ function HeaderSection() {
           <img src="img/openrefine_logo.svg" alt="OpenRefine logo" style={{maxWidth: "min(250px, 20vw)", display: 'block', margin: 'auto'}}/>
         </div>
       </div>
+     <div>
+      <Link className="button button--primary" to="https://forms.gle/EAqF1uYyDNsoaPyz8" style={{fontSize: '1.5em'}}>💎 Participate to the 2026 user survey ✨</Link>
```
tfmorris:
```suggestion
      <Link className="button button--primary" to="https://forms.gle/EAqF1uYyDNsoaPyz8" style={{fontSize: '1.5em'}}>💎 Participate in the 2026 user survey ✨</Link>
```

tfmorris (changes requested):
One small grammar fix for the link label, but otherwise looks OK. I think the banner might work better at the top of the page, rather than buried in the middle, but I'm not a designer.

Not sure how to test the `events.json`, but we can fix it up later if there's an issue.

From a process point of view, it seems more like something that the project management committee needs to approve rather than the core developers, but I'm fine with reviewing too.

